PR #14, opened 2019-07-13
Changing the readme

ki3ranoday:


src/_main_/settings.py
```diff
@@ -30,6 +30,8 @@
 DEBUG = True
```
Opoku-Agyemang: Are you sure this is the latest version of this file?  It seems some or all of the changes here are already done in the adding-user-portal-views branch.  Or did you branch directly from master?

Opoku-Agyemang:
LGTM

PR #115, opened 2020-04-24
Enable carbon-reduction targets and attained values

BradHN1:
The API includes carbon_footprint_reduction target and attained values the same as it does households and actions completed.  This allows the community portal to show carbon_footprint_reduction graphs (enabled by a community admin by setting target > 0). Also include the attained carbon_footprint_reduction for teams.

Addresses API#110 and Frontend-portal#186

### DIFF
--- a/src/api/store/team.py
+++ b/src/api/store/team.py
@@ -39,7 +39,7 @@ class TeamStore:
       teams = Team.objects.filter(community=community)
       ans = []
       for team in teams:
-        res = {"members": 0, "households": 0, "actions": 0, "actions_completed": 0, "actions_todo": 0}
+        res = {"members": 0, "households": 0, "actions": 0, "actions_completed": 0, "actions_todo": 0, "carbon_footprint_reduction": 0}
         res["team"] = team.simple_json()
         # team.members deprecated
         # for m in team.members.all():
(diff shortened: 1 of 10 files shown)


src/api/store/team.py
```diff
@@ -50,8 +50,12 @@ def team_stats(self, context: Context, args) -> (list, MassEnergizeAPIError):
           res["households"] += user.real_estate_units.count()
           actions = user.useractionrel_set.all()
           res["actions"] += len(actions)
-          res["actions_completed"] += actions.filter(**{"status":"DONE"}).count()
+          done_actions = actions.filter(**{"status":"DONE"})
```
massenergize-admin, line 53: nit: probably better to do `actions.filter(status="DONE").count()`

src/api/urls.py
```diff
@@ -1,6 +1,6 @@
-from django.urls import path, re_path
-from django.conf.urls import url
-from .views import *
+#from django.urls import path, re_path
```
massenergize-admin, line 1: if we dont need these just delete

src/database/models.py
```diff
@@ -195,9 +195,8 @@ class Goal(models.Model):
   more_info = JSONField(blank=True, null=True)
   is_deleted = models.BooleanField(default=False, blank=True)
 
-
-  def get_status(self):
-    return CHOICES["GOAL_STATUS"][self.status]
+  #def get_status(self):
```
massenergize-admin, line 198: same here remove all together since wee dont use

PR #123, opened 2020-05-07
Fix for two issues - deleting users and deleted actions

BradHN1:
Addresses the following issues:
API#121 - Fix problem with deleted actions causing crashes (Admin#159)
API#122 - Deleting users (Admin

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,8 @@ timelog.csv
 \.vscode/
 
 *.env
+
+src/*.jpg
+src/*.jpeg
+src/*.png
+
--- a/src/api/store/action.py
+++ b/src/api/store/action.py
@@ -12,8 +12,14 @@ class ActionStore:
 
   def get_action_info(self, context: Context, action_id) -> (dict, MassEnergizeAPIError):
     try:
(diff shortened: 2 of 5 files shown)


src/api/store/action.py
```diff
@@ -24,15 +30,19 @@ def list_actions(self, context: Context,community_id, subdomain) -> (list, MassE
     try:
       actions = []
       if community_id:
-        actions = Action.objects.select_related('image', 'community').prefetch_related('tags', 'vendors').filter(community__id=community_id, is_deleted=False)
+        actions = Action.objects.select_related('image', 'community').prefetch_related('tags', 'vendors').filter(community__id=community_id)
```
massenergize-admin, line 33: nice!

.gitignore
```diff
@@ -8,3 +8,8 @@ timelog.csv
 \.vscode/
 
 *.env
+
+src/*.jpg
```
massenergize-admin, line 12: where are these images coming from?

BradHN1:
> LGTM ✅
> I'm wondering why you have to add src/*.jpg to the gitignore. We should not have any pictures in the src right?

Only because of a bug in the carbon_calculator content importing which regularly copies my graphics files into the current directory, which I then delete.  It is about time I figure out how to fix that.

PR #154, opened 2020-06-19
Add team actions endpoint

jkatofsky:
I added an endpoint that calculates the completed actions, by category, for all the members of a given team, which will be called on the new single team page.

### DIFF
--- a/src/api/store/graph.py
+++ b/src/api/store/graph.py
@@ -100,16 +100,9 @@ class GraphStore:
       if not team:
         return None, InvalidResourceError()
 
-      #TODO: addresss the commented-out code below (was not letting me do calls in testing)
-      #TODO: fix miscounting:
-      # - for Wayland Happy Hollow 2, counting 15 instead of 16 total actions
-      # - for funny ones, counting 39 instead of 35 total actions
-
-      # print(context.is_prod, context.user_is_admin(), team.is_published)
-
-      # if context.is_prod and not context.user_is_admin():
-      #   if not team.is_published:
-      #     return None, CustomMassenergizeError("Content Not Available Yet")
+      if context.is_prod and not context.user_is_admin():
+        if not team.is_published:
+          return None, CustomMassenergizeError("Content Not Available Yet")
   
       members = TeamMember.objects.filter(team=team).all()
 
@@ -117,10 +110,9 @@ class GraphStore:
       for member in members:
         completed_action_rels.extend(member.user.useractionrel_set.filter(status="DONE").all())
 
-      prefetch_related_objects(completed_action_rels, "action__tags")
-
       categories = TagCollection.objects.get(name="Category").tag_set.all()
 
+      prefetch_related_objects(completed_action_rels, "action__tags")
       data = [
         {
           "id"   : category.id,
@@ -129,10 +121,6 @@ class GraphStore:
         }
         for category in sorted(categories, key=lambda category: category.name)
       ]
-
-      print("completed actions: %s\n" % completed_action_rels, "amount: %s\n" % len(completed_action_rels))
-      print("categories: %s\n" % categories)
-      print("data: %s\n" % data, "sum of values: %s\n" % (sum([entry["value"] for entry in data])))
 
       res = {
         "data": data,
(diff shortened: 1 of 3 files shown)


src/api/store/graph.py
```diff
@@ -87,6 +87,55 @@ def graph_actions_completed(self, context: Context, args) -> (Graph, MassEnergiz
...
+  
+      members = TeamMember.objects.filter(team=team).all()
+
+      completed_action_rels = []
+      for member in members:
+        completed_action_rels.extend(member.user.useractionrel_set.filter(status="DONE").all())
+
+      categories = TagCollection.objects.get(name="Category").tag_set.all()
+
+      prefetch_related_objects(completed_action_rels, "action__tags")
```
Opoku-Agyemang, line 115: i think you can combine line 113 and 115 into something like 
```
TagCollection.objects.prefetch_related_objects("action__tags").tag_set.all()
```

src/api/store/graph.py
```diff
@@ -87,6 +87,55 @@ def graph_actions_completed(self, context: Context, args) -> (Graph, MassEnergiz
...
+      res = {
+        "data": data,
+        "title": "Actions Completed By Members Of Team",
+        "team": team.info()
+      }
+   
+      return res, None
+        
+    except Exception as e:
+      import traceback
```
Opoku-Agyemang, line 134: is this for your own debugging?
Do you want to ship this?

src/api/store/graph.py
```diff
@@ -87,6 +87,55 @@ def graph_actions_completed(self, context: Context, args) -> (Graph, MassEnergiz
...
+        completed_action_rels.extend(member.user.useractionrel_set.filter(status="DONE").all())
+
+      categories = TagCollection.objects.get(name="Category").tag_set.all()
+
+      prefetch_related_objects(completed_action_rels, "action__tags")
+      data = [
+        {
+          "id"   : category.id,
+          "name" : category.name,
+          "value": len(list(filter(lambda action_rel : category in action_rel.action.tags.all(), completed_action_rels)))
```
Opoku-Agyemang, line 120: i think you can take advantage of the sql generation to do a sort by.  I think django let's you fetch with a sort by

src/api/store/graph.py
jkatofsky: I simply modelled this piece of code after what was done in the function above - `graph_actions_completed`, assumed it was desired

src/api/store/graph.py
jkatofsky: I can't figure out code that would combine these lines; I don't think it's possible to call `prefetch_related` on the `TagCollection` model, because `action__tags` (which is what I want to pre-fetch) is a field of `UserActionRel`, not `TagCollection`.

src/api/store/graph.py
jkatofsky: done - see most recent commit

src/api/store/graph.py
Opoku-Agyemang: yh i would like us to remove all those traces in favor of using a logging service.  for now you can keep it but we should be using a proper logging service for proper debugging.

src/api/store/graph.py
```diff
@@ -87,6 +87,55 @@ def graph_actions_completed(self, context: Context, args) -> (Graph, MassEnergiz
...
+        completed_action_rels.extend(member.user.useractionrel_set.filter(status="DONE").all())
+
+      categories = TagCollection.objects.get(name="Category").tag_set.order_by("name").all()
+
+      prefetch_related_objects(completed_action_rels, "action__tags")
+      data = [
+        {
+          "id"   : category.id,
+          "name" : category.name,
+          "value": len(list(filter(lambda action_rel : category in action_rel.action.tags.all(), completed_action_rels)))
```
Opoku-Agyemang, line 120: i love the mastery and clean use of list comprehension here.  However i would go for more code comprehension here just for easier maintenance.  I can understand what you have done, but as we get more and more people to help out it might be better to break complex pieces like this into smaller chunks just for easier code readability.

Opoku-Agyemang: Also i think you dont need to do `list(` when you are using filter since it already returns an iterable.  so `len(filter...` would be more efficient.

jkatofsky: removed `all()` and made changes for readability. but as for removing the `list` cast, it causes `TypeError: object of type 'filter' has no len()`. While `filter` returns an interable, `len` takes in a sequence or collection, specifically.

jkatofsky:
> Happy to +1 it if you will add it later but we really need to avoid piling more tech debt in terms of tests

Sure - I will make sure to add tests later. Now I have my hands full with the teams redesign and CAdmin data download feature, but I would gladly be involved in the general testing effort in an upcoming sprint.

PR #160, opened 2020-06-26
Deprecating vulnerable v2 routes 

massenergize-admin:
This PR fixes
* Broken Authentication 
- Administrative Privilege Escalation 
- High Lack of API Authentication 

Resolution: Deprecate V2 routes which did not have any layer of security


massenergize-admin:
> All good. Corresponding changes need in Frontend-portal?

Unless we are using v2/ routes there which we should not be.  I am going to be doing some further tests.

jkatofsky:
There appear to be multiple uses of v2 routes on the frontend portal. for example, my client is making these requests from the profile page:

- `https://api-dev.massenergize.org/v2/user/e/josh.katofsky@massenergize.org/actions?status=TODO`
- `https://api-dev.massenergize.org/v2/user/ed7b6a9e-4d7a-4326-8167-bfde218b1af0`

@massenergize-admin @BradHN1

PR #173, opened 2020-07-27
Team model enhancements

BradHN1:
This PR address a number of API changes that are needed for further front-end updates.
Among the issues addressed are:
API 106, 109, 116, 144, 145, 146


BradHN1:
Migrations seem correct, which I've applied locally.  Need to talk with @sam about getting it merged

src/_main_/settings.py
```diff
@@ -135,21 +135,21 @@
...
+    'remote-default': {
         'ENGINE': os.environ.get('DATABASE_ENGINE'),
         'NAME': os.environ.get('DATABASE_NAME'),
         'USER': os.environ.get('DATABASE_USER'),
         'PASSWORD': os.environ.get('DATABASE_PASSWORD'),
         'HOST': os.environ.get('DATABASE_HOST'),
         'PORT': os.environ.get('DATABASE_PORT')
     },
-    'local-default': {
+    'default': {
```
massenergize-admin, line 146: please switch

src/_main_/settings.py
```diff
@@ -135,21 +135,21 @@
...
         'ENGINE': os.environ.get('DATABASE_ENGINE'),
         'NAME': 'postgres2',
         'USER': 'Brad',
         'PASSWORD': '',
         'HOST': 'localhost',
-        'PORT': '5432'
+        'PORT': '5555'
```
massenergize-admin, line 152: i would love if this was in the env file so everyone can put their port. And that goes for all the variables above.  
Its probably better if your own local version of `dev.env` sets the db values to these and so the default one relies on that

src/database/models.py
```diff
@@ -571,14 +575,24 @@ class Team(models.Model):
...
+
   admins = models.ManyToManyField(UserProfile, related_name='team_admins', 
     blank=True) 
   members = models.ManyToManyField(UserProfile, related_name='team_members', 
     blank=True) 
+
+  # may change this from ForeignKey to ManyToManyField to allow team to span communities
   community = models.ForeignKey(Community, on_delete=models.CASCADE)
+  image = models.ManyToManyField(Media, related_name='team_image')                         # 0 or more photos - could be a slide show
+  video = models.ForeignKey(Media, on_delete=models.SET_NULL, related_name='team_video', null=True)                               # allow one video
```
massenergize-admin, line 591: i dont know but it think its always a better idea for people to upload videos on youtube or vimeo and then insert a link.  I strongly advise against storing "video" files ourselves.

src/database/models.py
```diff
@@ -1971,168 +2041,69 @@ class Meta:
     verbose_name_plural = "HomePageSettings"
 
 
-class ActionsPageSettings(models.Model):
```
massenergize-admin, line 1974: i would not delete any of these per se.  I made all these pages different because they are mostly all not the same.  This will take off some pages and force us into doing backfills which are unnecessary in my honest opinion.

src/_main_/settings.py
BradHN1: I added IS_LOCAL and a local.env file - same as dev.env but with my database settings.

src/_main_/settings.py
BradHN1: done

src/database/models.py
BradHN1: Good point.  Changed to video_link

src/database/models.py
BradHN1: I'm not sure whatyou meant, it doesn't take off any pages or require backfills.  In fact all the page settings models except HomePageSettings were identical, except for DonatePageSettings which had one additional field; I added that. I wanted to take advantage of the commonality of page settings to add the optional social media links and images

jkatofsky:
@BradHN1 we will need a `teams_page_settings` route in order for CAdmins to change those settings and for the community portal to access those settings.

BradHN1:
Ok - do you want to write that route?  I could take a crack at it this evening.

- Brad

++++++++++++++++++++
Bradley Hubbard-Nelson
221 Nashawtuc Rd
Concord, MA 01742
bradhn@mindspring.com
hubbardnelson.org <http://hubbardnelson.org/>
Tel: 978-505-7128 (cell)

jkatofsky:
I have some logistics I need to take care of for the rest of today, so won't be able to do this today. Feel free to take a crack; it seems like you just have to copy over the logic for the other page settings routes. Otherwise, I'll work on this first thing tomorrow.

jkatofsky:
I noticed that you haven't changed any of the methods in the team store except for `create`. However, when someone joins/leaves a sub-team, should that not also create/delete a `TeamMember` of the parent team?

Or, alternatively, when accessing the `info` of a parent team or calling `list` or `stats` we will need to loop over TeamMembers for all of the sub-teams since they should also count as memebrs of the parent team.

PR #198, opened 2020-09-24
Bug fix: ensure parameters like vendor ID for testimonials passed as integers, not strings

BradHN1:


Opoku-Agyemang (commented):
I don’t see how your pr addresses any issue with Boolean values.  Am I missing something ?

BradHN1:
@Opoku-Agyemang - I observed that these context values is_prod, is_dev and is_sandbox were passed as strings, like "false" with small "f"; if you test the string value: if "false" returns True.  The parse_bool fixes that.  I've tested it and it with this change those values are interpreted correctly.  Josh also observed this error, that those values were coming out wrong.

Opoku-Agyemang (approved):

PR #202, opened 2020-10-02
Address community portal issue 318 and 244

BradHN1:
Fixes to team creation and updating and subscribe to newsletter

### DIFF
--- a/src/api/handlers/team.py
+++ b/src/api/handlers/team.py
@@ -7,7 +7,7 @@ from _main_.utils.massenergize_response import MassenergizeResponse
 from types import FunctionType as function
 from _main_.utils.context import Context
 from _main_.utils.validator import Validator
-from _main_.utils.common import parse_str_list
+from _main_.utils.common import parse_str_list, parse_bool
 from api.decorators import admins_only, super_admins_only, login_required
 
 class TeamHandler(RouteHandler):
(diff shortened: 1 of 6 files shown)


src/api/handlers/team.py
```diff
@@ -58,6 +58,11 @@ def create(self, request):
 
     admin_emails = args.pop('admin_emails', '')
     args["admin_emails"] = parse_str_list(admin_emails)
+
+    parentId = args.pop('parent_id', None)
+    if parentId and parentId != 'undefined':
```
massenergize-admin, line 63: this is great! Do you mind turning this into a helper function like `is_empty()` where it checks if its `undefined` `null` `None` etc

src/api/store/team.py
```diff
@@ -155,12 +155,14 @@ def create_team(self, context:Context, args) -> (dict, MassEnergizeAPIError):
 
       # TODO: this code does will not make sense when there are multiple communities for the team...
       # TODO: create a rich email template for this?
-      is_published = context.user_is_admin()
+      #is_published = context.user_is_admin()
```
massenergize-admin, line 158: remove if not needed

Opoku-Agyemang (approved):

src/api/store/team.py
```diff
@@ -213,12 +217,16 @@ def update_team(self, team_id, args) -> (dict, MassEnergizeAPIError):
             team.parent = parent
         
         if logo:
-          logo = Media.objects.create(file=logo, name=f"{slugify(team.name)}-TeamLogo")
-          logo.save()
-          team.logo = logo
+          # if existing logo, the string length is around 300 characters
+          # If a new logo updated, this will be the length of the file, much larger than that       
+          new_logo = len(logo) > 1000
```
Opoku-Agyemang, line 222: i am confused.  the comment talks of 300 chars and the code dictates 1000

BradHN1: That number is arbitrary.  A real image file will be at least a few kilobytes, much larger than 1000.  So 1000 seems like a safe cutoff to test for a photo upload.

PR #209, opened 2020-10-15
Fix subscriber bug

BradHN1:
Portal issue 306
Also add some carbon calculator unit tests Tori (HS intern) created

### DIFF
--- a/src/carbon_calculator/tests/tests.py
+++ b/src/carbon_calculator/tests/tests.py
@@ -189,15 +189,6 @@ class CarbonCalculatorTest(TestCase):
         else:
             print("carbon_calculator results consistent with input data from "+str(oldtime))
 
-def outputLine(data, filename, new=False):
-    tag = "a"
-    if new:
-        tag = "w"
-
-    f = open(filename, tag)
-    f.write(str(data) + "\n")
-    f.close()
-    # 1 !!!!!!! Works !!!!!!!
     def test_info_events(self):
         ''' Tests /cc/info/events url and returns a json of events. '''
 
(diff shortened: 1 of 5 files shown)


BradHN1 (commented):
My changes are ok, I've tested them

src/carbon_calculator/tests/tests.py
```diff
@@ -183,6 +198,155 @@ def pretty_print_diffs(self, diffs, oldtime):
         else:
             print("carbon_calculator results consistent with input data from "+str(oldtime))
 
+    def test_info_events(self):
+        ''' Tests /cc/info/events url and returns a json of events. '''
+
+        event_list = self.client.get("/cc/info/events", {}) # status code = 200
+        event_list_json = jsons.loads(event_list.content)   # loads into json
+        #print(event_list_json)                             # uncomment this to see json
```
Opoku-Agyemang, line 206: So i see a bunch of these commented out print statements.  Ideally, we don't really want "dead code" in there.  If its not a comment explaining something / some complex code we don't want to rewrite then it should probably not be there.  

One trick we could use here is that, just like we have DEBUG=True/False in settings.py, you could declare a similar constant in this file (maybe called `DEBUG=True` and then put all these print statements under an if statement that checks if this DEBUG=True then it prints else it leaves it.  So you won't need to uncomment anything if you wanted those print statements.  You can control it with just one change.

src/carbon_calculator/tests/tests.py
```diff
@@ -183,6 +198,155 @@ def pretty_print_diffs(self, diffs, oldtime):
...
+        #print(field_value)
+
+        event_url = "/cc/info/group/" + field_value
+        #print(event_url)
+        event_info = self.client.get(event_url, {})
+        event_json = jsons.loads(event_info.content)
+        #print(event_json)
+
+    # 8 !!!!!!! WORKS !!!!!!!
+    def test_info_stations_one_station(self):
```
Opoku-Agyemang, line 255: nice! i like the naming of this ✅

src/carbon_calculator/tests/tests.py
```diff
@@ -183,6 +198,155 @@ def pretty_print_diffs(self, diffs, oldtime):
...
+        #print(event_json)
+
+    #extra
+    def test_get_action_list(self):
+        impact_info = self.client.get("/cc/info/actions", {})
+        #print("actions:")
+        #print(jsons.loads(impact_info.content))
+
+    # 12 !!!!!! WORKS !!!!!!!
+    def test_estimate_actions(self):
```
Opoku-Agyemang, line 274: is this testing actions that are valid? could you test invalid actions too?

PR #216, opened 2020-10-29
API fixes - Issues 214, 215

BradHN1:


src/api/handlers/download.py
```diff
@@ -76,7 +76,7 @@ def teams_download(self, request):
     context: Context = request.context
     args: dict = context.args
     community_id = args.pop('community_id', None)
-    teams_data, community_name, err = self.service.teams_download(context, community_id)
+    (teams_data, community_name), err = self.service.teams_download(context, community_id)
```
massenergize-admin, line 79: There is an issue here.  And its not your fault.  
1. The issue starts from `store/download.py` where all the return statements are inconsistent.  Can you make them consistent?  as in, either always return two objects or three.  For now its different for different cases.  Good catch.
2. Whatever you do for the store layer, treat the service and handler layer respectively.

BradHN1: @massenergize-admin : 
I've reviewed the download.py in store, services and handlers, and as I see it they always return 2 objects:
result (a tuple of 2 elements:  data and community_id) or None if is an error;
error (usually None, unless there is an error)

So I think they are already consistent, with the change I made.  They weren't consistent before my change

Opoku-Agyemang: i think you're right.  That said, i am definitely not a fun of 
`(self._community_teams_download(community.id), community.name), None`

you might as well return three things or a dictionary with two keys or something.

Opoku-Agyemang (approved):
LGTM ✅

PR #219, opened 2020-11-01
Fix admin issues 218 and 253

BradHN1:
Fixes two broken features:
All communities download
Create Community

### DIFF
--- a/src/api/store/community.py
+++ b/src/api/store/community.py
@@ -181,31 +181,64 @@ class CommunityStore:
           comm_admin.members.add(owner)
           comm_admin.save()
 
-      #also clone all global actions for this community
-      #global_actions = Action.objects.filter(is_global=True)
-      #for action_to_copy in global_actions:
-      #  print(action_to_copy)
-      #  action_to_copy_id = action_to_copy.id
-      #  old_tags = action_to_copy.tags.all()
-      #  old_vendors = action_to_copy.vendors.all()
-      #  new_action: Action = action_to_copy
-      #  new_action.pk = None
-      #  new_action.is_published = False
-      #  new_action.is_global = False
+      # Also clone all template actions for this community
+      # 11/1/20 BHN: Add protection against excessive copying in case of too many actions marked as template
+      # Also don't copy the ones marked as deleted!
+      global_actions = Action.objects.filter(is_deleted=False, is_global=True)
+      count = global_actions.count()
+      MAX_TEMPLATE_ACTIONS = 25
+      num_copied = 0
 
-        #make sure the action title does not exceed the max length expected
-      #  suffix = f'-{new_community.subdomain}-{action_to_copy_id}'
-      #  new_action.title = new_action.title[:SHORT_STR_LEN-len(suffix)]+suffix 
+      for action_to_copy in global_actions:
+        action_to_copy_id = action_to_copy.id
+        old_tags = action_to_copy.tags.all()
+        old_vendors = action_to_copy.vendors.all()
+        new_action: Action = action_to_copy
+        new_action.pk = None
+        new_action.is_published = False
+        new_action.is_global = False
+
+        # remove the "TEMPLATE", "TEMP" or "TMP" prefix or suffix
+        old_title = new_action.title
+        new_title = old_title
+
+        loc = old_title.find("TEMPLATE")
+        if loc == 0 :
+          new_title = old_title[9:]
+        elif loc > 0:
+          new_title = old_title[0:loc-1]
+        else:
+          loc = old_title.find("TEMP")
+          if loc==0:
+            new_title = old_title[5:]
+          elif loc>0:
+            new_title = old_title[0:loc-1]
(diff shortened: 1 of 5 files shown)


src/api/store/community.py
```diff
@@ -183,29 +182,31 @@ def create_community(self,context: Context, args) -> (dict, MassEnergizeAPIError
...
-      global_actions = Action.objects.filter(is_global=True)
-      for action_to_copy in global_actions:
-        action_to_copy_id = action_to_copy.id
-        old_tags = action_to_copy.tags.all()
-        old_vendors = action_to_copy.vendors.all()
-        new_action: Action = action_to_copy
-        new_action.pk = None
-        new_action.is_published = False
-        new_action.is_global = False
+      #global_actions = Action.objects.filter(is_global=True)
```
Opoku-Agyemang, line 185: could you explain why you are removing all this code?

src/api/store/community.py
```diff
@@ -183,29 +182,31 @@ def create_community(self,context: Context, args) -> (dict, MassEnergizeAPIError
...
+      #  old_vendors = action_to_copy.vendors.all()
+      #  new_action: Action = action_to_copy
+      #  new_action.pk = None
+      #  new_action.is_published = False
+      #  new_action.is_global = False
 
         #make sure the action title does not exceed the max length expected
-        suffix = f'-{new_community.subdomain}-{action_to_copy_id}'
-        new_action.title = new_action.title[:SHORT_STR_LEN-len(suffix)]+suffix 
+      #  suffix = f'-{new_community.subdomain}-{action_to_copy_id}'
```
Opoku-Agyemang, line 197: again why are you removing?  i am expecting at least a comment

BradHN1: Now that we don't need to have unique action names, there is no reason to add the action_to_copy_id.  It is just a pain for community admins and takes away from the useful length of the name.

src/api/store/community.py
BradHN1: Thank you for asking! What I saw was that in creating a community there were hundreds of actions being copied, and that is a problem!  It turned out that it wasn't filtering for is_deleted=False so there was a bunch of crap copied for each community that was not showing up and taking a super long time.

I've reinstated the cloning and improved the naming to be more convenient for community admins

src/api/store/community.py
```diff
@@ -193,19 +198,48 @@ def create_community(self,context: Context, args) -> (dict, MassEnergizeAPIError
         new_action.is_published = False
         new_action.is_global = False
 
-        #make sure the action title does not exceed the max length expected
-        suffix = f'-{new_community.subdomain}-{action_to_copy_id}'
-        new_action.title = new_action.title[:SHORT_STR_LEN-len(suffix)]+suffix 
+        # remove the "TEMPLATE", "TEMP" or "TMP" prefix or suffix
+        old_title = new_action.title
```
Opoku-Agyemang, line 202: what is the goal of this?

src/api/store/community.py
```diff
@@ -193,19 +198,48 @@ def create_community(self,context: Context, args) -> (dict, MassEnergizeAPIError
...
         # if(Action.objects.filter(title=new_action.title, community=new_community).count() == 0):
         new_action.save()
-
         new_action.tags.set(old_tags)
         new_action.vendors.set(old_vendors)
 
         new_action.community = new_community
         new_action.save()
+        num_copied += 1
+        if num_copied >= MAX_TEMPLATE_ACTIONS:
```
Opoku-Agyemang, line 240: i think we should not do this.  
I think there is another way to go about what you want to do.  I did the same for Dev.

Basically, in the past there was a bug where while creating a community, in the process of copying the template actions, we were not setting those new actions being created to be is_template=False and so the next time we created something again, those actions showed up again.  It was quite funny when i saw it.  

What i did for dev, which we should do for canary, is to do a backfill instead.  Use the backfill route to delete all nonsense action templates and literally wipe them out.  

And since the bug has been fixed we wont have to deal with that nuisance of creating nonsense actions anymore.  

Good catch though.  I think this the best way to go about it and not necessarily set a bound for number of actions.

src/api/store/community.py
BradHN1: The goal of this is to have the title of the new action be simple, just the name of the template action with "-Concord" on the end.  Makes the cadmin's job easier and there is no reason to have "TEMPLATE" as part of the name or "Copy XXXX".  The comment explains that.

src/api/store/community.py
BradHN1: On the API to just blindly copy hundreds of actions, which is what it was doing, is what I call kamakazi code; leave yourself open to filling the database with junk and hanging the API up in an endless loop.  That is what was happening so this is the fix.  A bound on the number of template actions is a good idea in my opinion; there are 11 now and I could imagine up to 20 but not more than 25.  However- as we get into more regions we may consider having different groups of template actions and at that time it would need to change.

PR #223, opened 2020-11-17
Changes for CarbonSaver to use UserProfile

BradHN1:
Would like to get review on these changes


src/api/handlers/userprofile.py
```diff
@@ -55,8 +55,8 @@ def info(self, request):
   def create(self, request):
     context: Context = request.context
     args: dict = context.args
-    validator: Validator = Validator()
-    args, err = (validator
+    #validator: Validator = Validator()
```
Opoku-Agyemang, line 58: nice!  just remove it then since its not needed

src/carbon_calculator/carbonCalculator.py
```diff
@@ -43,20 +46,32 @@ def SavePic2Media(picURL):
...
             return None
         else:
-            file_name =  "carbon_calculator/assets/" + picURL.split("/")[-1]  # There's probably a better way of doing this but this is just a quick example
-            fp = open(file_name,"wb")
-            fp.write(resp.content)
-            fp.close()
-            media = CarbonCalculatorMedia(file = file_name)
-            #media = CarbonCalculatorMedia.objects.create(file=file_name)
+            image = resp.content
+            file_name =  picURL.split("/")[-1]
```
Opoku-Agyemang, line 50: nit: we should probably make sure the picUrl is not empty.

src/carbon_calculator/carbonCalculator.py
```diff
@@ -43,20 +46,32 @@ def SavePic2Media(picURL):
...
+            file_type_ext = file_name.split(".")[-1]
```
Opoku-Agyemang, line 51: we should probably make sure that it contains `.` inside first or at least check that `file_type_ext` is not empty before proceeding

src/carbon_calculator/carbonCalculator.py
```diff
@@ -43,20 +46,32 @@ def SavePic2Media(picURL):
...
+                content_type = 'image/png'
+
+            # Create a new Django file-like object to be used in models as ImageField using
+            # InMemoryUploadedFile.  If you look at the source in Django, a
+            # SimpleUploadedFile is essentially instantiated similarly to what is shown here
+            #image_file = InMemoryUploadedFile(img_io, None, 'foo.jpg', 'image/jpeg',
+            image_file = InMemoryUploadedFile(img_io, None, file_name, content_type,
+                                  None, None)
+
+            media = CarbonCalculatorMedia.objects.create(file=image_file, name=f"{slugify(file_name)}")
```
Opoku-Agyemang, line 66: i am a bit confused.  i think you can just pass in image here and it will be fine.  At this point it should already be in memory so no need to redeclare an InMememoryUploadedFile

src/carbon_calculator/models.py
```diff
@@ -180,51 +187,51 @@ class Meta:
...
+#    """
+#    A class used to represent a Calculator User
 #
-    class Meta:
-        db_table = 'user_profiles_cc' 
+#    Note: Authentication is handled by firebase so we just need emails
+#
+#   Attributes
+#    ----------
+#    email : str
```
Opoku-Agyemang, line 198: if this code is dead lets probably remove it?

Opoku-Agyemang (commented):
LGTM but a few nit picky comments.

src/carbon_calculator/carbonCalculator.py
BradHN1: sure

src/carbon_calculator/carbonCalculator.py
BradHN1: There may be other ways to do it, but I tried a lot of things that didn't work including passing the image in directly.  This came from an example I found, and works for both png and jpg.  it was important to specify the content_type for the in memory file.

src/carbon_calculator/models.py
BradHN1: fine

src/carbon_calculator/carbonCalculator.py
BradHN1: ok, looks safe to me since inside try, but I added a check

Opoku-Agyemang (approved):

PR #232, opened 2021-02-11
Handle Update cases where community set to "no community" etc

BradHN1:
On updating Actions, Testimonials, Events, etc, enable clearing the link to primary community, logo, etc

This goes along with the Admin Portal changes to add "--" to dropdowns


Opoku-Agyemang (approved):
I think setting things to none is redundant here but it’s inconsequential so you can go ahead

PR #234, opened 2021-02-16
Fix community accounting by defining geographic communities

BradHN1:
This is the implementation of geographic communities via zipcodes, cities, counties, states, countries

Documented in [Community Membership and Accounting](https://docs.google.com/document/d/1yaTFLet1PJW6qXxVON_B2K6--Y0yDSti_9GktLNJikk/edit?usp=sharing)

For discussion, this is a draft


src/api/handlers/action.py
```diff
@@ -150,7 +150,7 @@ def community_admin_list(self, request):
   @super_admins_only
   def super_admin_list(self, request): 
     context: Context = request.context
-    args: dict = context.args
+    #args: dict = context.args
```
Opoku-Agyemang, line 153: nit: if not needed remove

src/api/handlers/event.py
```diff
@@ -181,7 +181,7 @@ def community_admin_list(self, request):
   @super_admins_only
   def super_admin_list(self, request):
     context: Context = request.context
-    args: dict = context.args
+    #args: dict = context.args
```
Opoku-Agyemang, line 184: same here

src/api/handlers/page_settings__home.py
```diff
@@ -18,7 +18,7 @@ def __init__(self):
 
   def registerRoutes(self):
     self.add("/home_page_settings.info", self.info) 
-    self.add("/home_page_settings.publish", self.info) 
+    #self.add("/home_page_settings.publish", self.publish)
```
Opoku-Agyemang, line 21: if not needed remove.  why are we commenting it out though?

src/api/handlers/page_settings__home.py
```diff
@@ -42,15 +42,16 @@ def info(self, request):
...
-    context: Context = request.context
-    args: dict = context.args
-    home_page_id = args.pop('home_page_id', None)
-    home_page_setting_info, err = self.service.get_home_page_setting_publish(home_page_id)
-    if err:
-      return MassenergizeResponse(error=str(err), status=err.status)
-    return MassenergizeResponse(data=home_page_setting_info)
+  # this route isn't used - so not sure whether it works
+  #@admins_only
+  #def publish(self, request):
```
Opoku-Agyemang, line 47: it is used in the admin.  on the community profile page

src/api/handlers/vendor.py
```diff
@@ -45,16 +45,17 @@ def info(self, request):
...
-    args: dict = context.args
-    args = rename_field(args, 'vendor_id', 'id')
-    args['is_published'] =True
-    vendor_info, err = self.service.update(args)
-    if err:
-      return MassenergizeResponse(error=str(err), status=err.status)
-    return MassenergizeResponse(data=vendor_info)
+  # this isn't used
+  # @admins_only
+  #def publish(self, request):
```
Opoku-Agyemang, line 50: if not needed remove

src/api/store/community.py
```diff
@@ -1,16 +1,206 @@
...
+from .utils import get_community_or_die, get_user_or_die, get_new_title, is_reu_in_community
+import random
+import zipcodes
 from sentry_sdk import capture_message, capture_exception
 
 class CommunityStore:
   def __init__(self):
     self.name = "Community Store/DB"
 
+  # QUESTION FOR SAM:
```
Opoku-Agyemang, line 15: if they are only used here and will only ever be used here then leave as is.  otherwise declaring them in a common package might be better

src/api/store/community.py
```diff
@@ -1,16 +1,206 @@
...
+  # These helper routines don't make use of self, should they be member functions as currently implemented, 
+  # or outside the CommunityStore class?
+
+  def _check_geography_unique(self, community, geography_type, loc):
```
Opoku-Agyemang, line 19: nit: maybe `geography_is_unique` is better.  Also for function specifications use """ .....""" instead of #

src/api/store/community.py
```diff
@@ -1,16 +1,206 @@
...
+    # ensure that this zip code is not part of another geographic zip code based community
+    check_communities = Community.objects.filter(is_geographically_focused=True, geography_type=geography_type, is_deleted=False).prefetch_related('location_set')
```
Opoku-Agyemang, line 21: nit: maybe `communities_to_check`

src/api/store/community.py
```diff
@@ -1,16 +1,206 @@
...
+    for check_community in check_communities:
```
Opoku-Agyemang, line 22: yeaah this is probably not wrong, but very expensive

src/api/store/community.py
```diff
@@ -1,16 +1,206 @@
...
+
+  def _update_locations(self, geography_type, location_set, community):
+        
+    # clean up in case there is garbage in there
+    if community.location_set:
+      community.location_set.clear()
+
+    # this is a list of zipcodes, towns, cities, counties, states        
+    location_list = location_set.replace(", ",",").split(",")  # passed as comma separated list
+    print("Community includes the following locations :"+str(location_list))
```
Opoku-Agyemang, line 56: remove print statement

src/api/store/community.py
```diff
@@ -1,16 +1,206 @@
...
+          # assume this is a town, see we can find the zip codes associated with it
+          ss = location.split('-')
+          town = ss[0]
+          if len(ss)==2:
+            state = ss[1]
+          else:
+            state = 'MA'
+            
+          zips = zipcodes.filter_by(city=town, state=state, zip_code_type='STANDARD' )
+          print("Number of zipcodes = "+str(len(zips)))
```
Opoku-Agyemang, line 87: remove print

src/api/store/community.py
```diff
@@ -1,16 +1,206 @@
...
+          if len(zips)>0:
+            for zip in zips:
+              print(zip)
```
Opoku-Agyemang, line 90: remove print

src/api/store/community.py
```diff
@@ -1,16 +1,206 @@
...
+          town = ss[0]
+          if len(ss)==2:
+            state = ss[1]
+          else:
+            state = 'MA'
+            
+          zips = zipcodes.filter_by(city=town, state=state, zip_code_type='STANDARD' )
+          print("Number of zipcodes = "+str(len(zips)))
+          if len(zips)>0:
+            for zip in zips:
```
Opoku-Agyemang, line 89: very expensive move

src/api/store/community.py
```diff
@@ -1,16 +1,206 @@
...
+      elif geography_type == 'CITY':
+        # check that this city is found in the zipcodes list
+        ss = location.split('-')
+        city = ss[0]
+        if len(ss)==2:
+          state = ss[1]
+        else:
+          state = 'MA'
+
+        zips = zipcodes.filter_by(city=city, state=state, zip_code_type='STANDARD' )
```
Opoku-Agyemang, line 112: same

src/api/store/community.py
```diff
@@ -1,16 +1,206 @@
...
+            print("City "+city+" found")
+
+        else:
+          print("No zipcodes found corresponding to city "+ city + ", " + state)
+          raise Exception("No zipcodes found corresponding to city "+ city + ", " + state)
+
+        self._check_geography_unique(community, geography_type, city)
+      elif geography_type == 'COUNTY':
+       # check that this county is found in the zipcodes list
+        ss = location.split('-')
```
Opoku-Agyemang, line 128: tbh this function seems weirdly long.  could you break it into chunks?  for instance all the logic in this elif could be put under a function called `county....`

Opoku-Agyemang (changes requested):
It is a lot of changes and should probably be broken into chunks.
1. I see that in the same PR u update some packages.  maybe let's make a separate PR for that.  If we could split this into a few PRs that will be great and better for review.
2. let's add some tests
3. see code for more comments

src/api/store/community.py
BradHN1: maybe you like another name better, but that is style, not substance.

src/api/store/community.py
BradHN1: No it isn't wrong, it is correct and I've checked it.  There might be ways of speeding up,  but this routine is only used when a super admin from the admin portal creates or modifies a community.  It runs quite quickly so I'm not at all worried about the computation time.   Unless you mean expensive in $$ and then I'd like you to explain why.

src/api/store/community.py
BradHN1: The print statements are important for me to validate that nothing goes wrong in the deployment to databases.  I don't think they do any harm but if you insist I will go back and eliminate them once the databases are migrated and the community definitions set.

src/api/store/community.py
BradHN1: Same comment.  I rely on the print for the migration and community definition setting on dev and prod DBs.

src/api/store/community.py
BradHN1: You've got to be kidding me.  This isn't doing database access, it is accessing an array.  Tell me why this is expensive.  Again, this is run by a Super admin once in a blue moon.

src/api/handlers/action.py
BradHN1: This comment is about style, not substance.  Can save time by ignoring stuff that isn't important.

src/api/handlers/page_settings__home.py
BradHN1: No, I do not think it is used.  communities.update is used, and works fine for that purpose.  I don't see a reason to keep an additional routine to be maintained for the same purpose.

src/api/store/community.py
BradHN1: This comment is about style, not substance.

src/api/store/userprofile.py
```diff
@@ -37,8 +38,8 @@ def _has_access(self, context: Context, user_id=None, email=None):
 
   def get_user_info(self, context: Context, args) -> (dict, MassEnergizeAPIError):
     try:
-      email = args.get('email', None)
-      user_id = args.get('user_id', None)
+      #email = args.get('email', None)
```
Opoku-Agyemang, line 41: why are you commenting these out?

src/api/store/userprofile.py
```diff
@@ -67,20 +69,64 @@ def add_household(self, context: Context, args) -> (dict, MassEnergizeAPIError):
...
+        address = json.loads(address)
+        street = address.get('street', '')
+        unit_number = address.get('unit_number', '')
+        zipcode = address.get('zipcode', '')
+        city = address.get('city', '')
+        county = address.get('county', '')
+        state = address.get('state', '')
+      else:
+        # get address from location string
+        loc_parts = location.capitalize().replace(" ","").split(',')
```
Opoku-Agyemang, line 86: i really really think you should have a helper function defined somewhere that does this work.  I could see other functions needed this.

src/api/store/community.py
```diff
@@ -1,16 +1,208 @@
...
+      
+        self._check_geography_unique(community, geography_type, location)
+
+      else:
+        raise Exception("Unexpected geography type: " + str(geography_type))
+      # should be a five character string
+      community.location_set.add(loc)
+
+  def _update_real_estate_units(self, community):
+    """
```
Opoku-Agyemang, line 177: Could you add some more context in the docstring on what this function does and why it is needed?

src/api/store/community.py
```diff
@@ -1,16 +1,208 @@
...
+          raise Exception("No zipcodes found corresponding to country "+ location)
+      
+        self._check_geography_unique(community, geography_type, location)
+
+      else:
+        raise Exception("Unexpected geography type: " + str(geography_type))
+      # should be a five character string
+      community.location_set.add(loc)
+
+  def _update_real_estate_units(self, community):
```
Opoku-Agyemang, line 176: can you call it `_update_real_estate_units_with_community`

Opoku-Agyemang (commented):
Much more clarity reading over this time:

Here are some standing questions
1. Are we creating unique locations or its possible to have duplicates ?
For example: if I create a geographic community and one of my zip codes is 100.
Later on Brad comes to register in that community with a household in zip code 100.  Do we create a new location for this user or we first find if a zip code exists in our locations already and associate that with the user?

Answer: it is not possible to have duplicates of zip codes.  When you create or update a community, the routine _check_geography_unique checks that no other zipcode based community includes that zipcode, and if it does raises an exception:         
===
if geography_type == 'ZIPCODE' and location.zipcode == loc:
          # zip code already used
          message = 'Zipcode %s is already part of another geographic community %s.' % (loc, check_community.name)
          print(message)
          raise Exception(message)
===

2. I see that we call `self._update_real_estate_units(new_community)` when we are creating a new geographic community.  Hypothetically, say we create a geographic location for Wayland and associate it with some zipcodes.  Later we create another geographic community for maybe the whole state, MA which will include wayland in there.  Would this not be overriding the community on all the wayland real estates to now be set to MA ?

Answer: there is a precedence, that if we create a  State based community, it does't override the smaller geographic communities within it.  At present, the larger community won't include the zipcodes or cities defined as communities within it in the counting, and to report on the totals we will add up the communities within it.  Aimee and I talked about this and added an issue 246 to address that in the future if we decide to.  That would require changes to the counting logic which would likely slow down the site for the larger communities.  In that case we were thinking of Boston and a zipcode based community within Boston.

src/database/models.py
```diff
@@ -260,8 +260,19 @@ class Community(models.Model):
...
   location = JSONField(blank=True, null=True)
+
+  # new - define the geographic area for a community (zipcodes, towns/cities, counties, states, countries)
+  geography_type = models.CharField(
+    max_length=TINY_STR_LEN,
+    choices=CHOICES.get("COMMUNITY_GEOGRAPHY_TYPES", {}).items(), 
+    blank=True, null=True)
+  
+  # location_set will define the range for geographic communities
+  location_set = models.ManyToManyField(Location, blank=True)
```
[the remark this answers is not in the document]
BradHN1: Good point.  Changed it to 'locations'

src/api/store/userprofile.py
BradHN1: Didn't need to.  It fixed the warning from the linter that those variables weren't used.  I think you had commented out the call to self._has_access, which used those variables, but since you left that comment in there I left the commented out variables in there too.

src/api/store/userprofile.py
BradHN1: Good point, and I think that showed a potential bug since it wasn't identical in the three places.  Now I've made a helper function split_location_string

src/api/store/community.py
BradHN1: done

src/api/store/community.py
BradHN1: OK

Opoku-Agyemang (approved):

PR #239, opened 2021-03-04
Django testing updates

BradHN1:
This is work in progress
This draft PR is to get feedback on how to set the user case properly

test_teams.py has method signInAs which isn't working yet - need some help to understand this

Want to use this as an example so an intern can work on all the API routes testing next month


kehillahghana (changes requested):
The way to get this working is not to bother about generating a firebase token at all.
Take a look here https://github.com/massenergize/api/blob/3139bf395c78e97dec84400dc00ebfda885e0a7c/src/api/services/auth.py#L48 

That is where we start creating the payload that we eventually encode as the token.  If i were you, i would write a helper method in a file called  test_helpers.py and in that file i will define a function that essentially constructs that payload and does this and encode it like we do here 
https://github.com/massenergize/api/blob/3139bf395c78e97dec84400dc00ebfda885e0a7c/src/api/services/auth.py#L57

So in your actual testing files, you can have a setUp function that goes about calling this helper method you created to get the cookie dynamically and you can set it in your requests..

Opoku-Agyemang (approved):

PR #249, opened 2021-03-23
Pongo backend development

frimpongopoku:
Included `more_info` to community model structure to be returned in json when community json comes to user portal


src/_main_/settings.py
```diff
@@ -24,7 +24,7 @@
 # ********  LOAD CONFIG DATA ***********#
 IS_PROD = False
 IS_CANARY = False
-IS_LOCAL = False
+IS_LOCAL = True
```
sopokuagyemang, line 27: flip it back

BradHN1: Don't commit settings.py unless there are real changes.

BradHN1 (approved):
OK

PR #251, opened 2021-04-03
Fix bugs from dev testing

BradHN1:
Fixed several bugs from prod testing of community accounting
Now ready for canary test


src/api/store/community.py
```diff
@@ -179,31 +217,56 @@ def _update_real_estate_units_with_community(self, community):
...
+
+      for reu in userProfile.real_estate_units.all():
+        if reu.address:
+          zip = reu.address.zipcode
+          if not isinstance(zip,str) or len(zip)!=5:
+            address_string = str(reu.address)
+            print("REU invalid zipcode: address = "+address_string)
+
+            # temporary fix
+            if address_string.find("carlisle")>=0:
```
Opoku-Agyemang, line 235: please let's not be doing hardcoded stuff here please.  Even if you need this in the mean time, i would put this in a dictionary file so you have a faster look up than chaining so many ifs.  Also put it in the utils.py

BradHN1: @Opoku-Agyemang   Agreed.  I moved all those cases into ZIPCODE_FIXES.json which won't get committed to github. Will slack it to you.

BradHN1:
Need to get this done and move to Canary.

Issue #146, opened 2020-06-02
Service Providers page - customizable settings by community


BradHN1:
The  service providers page should have customizable, graphics and text above the service providers list,
By default, default title, image and text would be used.
The settings should be customizable by community.

Suggest new model VendorsPageSettings like the other page settings classes

PR #254, opened 2021-04-06
Several bug fixes and issues addressed

BradHN1:
Issues addressed:
API #146 Service Providers page - customizable settings by community
API #242 Event names don't need to be unique

Portal #474 Community not using “Events” - menu item
Portal #537 Customize All Actions Page title and several more

Admin #307 Implement customizable Actions Page title and other pages


src/api/store/community.py
```diff
@@ -9,6 +13,26 @@
 import zipcodes
 from sentry_sdk import capture_message, capture_exception
 
+def _clone_page_settings(pageSettings, title, community):
+    """
+    Clone fht page settings for a new community from a template
```
Opoku-Agyemang, line 18: nit: what is fht?

BradHN1:
typo

src/api/store/community.py
```diff
@@ -340,88 +364,64 @@ def create_community(self,context: Context, args) -> (dict, MassEnergizeAPIError
...
-      if teamsPage:
-        teamsPage.pk = None 
-        teamsPage.title = f"Teams in this community"
-        teamsPage.community = new_community
-        teamsPage.is_template = False
-        teamsPage.save()
-    
-      admin_group_name  = f"{new_community.name}-{new_community.subdomain}-Admin-Group"
-      comm_admin: CommunityAdminGroup = CommunityAdminGroup.objects.create(name=admin_group_name, community=new_community)
+      print("Create admin group")
```
Opoku-Agyemang, line 422: let's remove.  If you need it temporarily just to for debugging or testing i think put a // TODO comment to remove afterwards so you can trace back to it.

src/api/store/community.py
```diff
@@ -506,14 +508,18 @@ def update_community(self, community_id, args) -> (dict, MassEnergizeAPIError):
...
 
-      #new_community = community.first()
       if logo:   
         cLogo = Media(file=logo, name=f"{args.get('name', '')} CommunityLogo")
         cLogo.save()
         community.logo = cLogo
         community.save()
 
       return community, None
+      if favicon:
```
Opoku-Agyemang, line 518: nice!

src/api/store/page_settings__aboutus.py
```diff
@@ -38,13 +38,23 @@ def create_about_us_page_setting(self, args) -> (dict, MassEnergizeAPIError):
...
+      page_image = args.pop('image', None)
+
       if about_us_page_id:
-        actions_page_setting = AboutUsPageSettings.objects.filter(id=about_us_page_id)
-        actions_page_setting.update(**args)
-        if not actions_page_setting:
+        page_setting = AboutUsPageSettings.objects.filter(id=about_us_page_id)
+        if not page_setting:
           return None, InvalidResourceError()
+        page_setting.update(**args)
```
Opoku-Agyemang, line 47: this will error out if there is anything in the args that is not consistent with a column in the database etc.  so to be sure its good to use, let's make sure the handler method cleans up the args real good.

src/api/store/community.py
BradHN1: a typo

src/api/store/community.py
BradHN1: Done

src/api/store/page_settings__aboutus.py
BradHN1: I agree with you, but here is a case of a lot of duplicated code.  The very same code is in page_settings__actions, page_settings__contactus, page_settings__donate, page_settings__teams, not of which do any cleaning up of args in the handler method.  
I was thinking of refactoring a bit since these models are basically the same, so the same code should check the args etc.

BradHN1:
Merging to deploy to dev

PR #268, opened 2021-05-15
Make goals consistent, in one place

BradHN1:
Also include Owen's test routines and bug fixes

### DIFF
--- a/src/api/handlers/testimonial.py
+++ b/src/api/handlers/testimonial.py
@@ -34,44 +34,40 @@ class TestimonialHandler(RouteHandler):
   def info(self, request):
     context: Context = request.context
     args: dict = context.args
-    args = rename_field(args, 'testimonial_id', 'id')
+
+    self.validator.expect("id", int, is_required=True)
+    self.validator.rename("testimonial_id", "id")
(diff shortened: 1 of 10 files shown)


src/api/handlers/testimonial.py
```diff
@@ -34,44 +34,40 @@ def registerRoutes(self):
...
-      # if user specifies other_vendor and passed to API - should record it as an unapproved vendor
-
-    is_approved = args.pop("is_approved", None)
-    if is_approved:
-      args["is_approved"] = parse_bool(is_approved)
-    is_published = args.get("is_published", None)
-    if is_published:
-      args["is_published"] = parse_bool(is_published)
+
+    self.validator.expect("title", str, is_required=True)
```
Opoku-Agyemang, line 56: btw you can chain the expects like
```
self.validator
  .expect("title", str, is_required=True)
  .expect('community', int) ......

```

Opoku-Agyemang (approved):

src/api/handlers/testimonial.py
BradHN1: @sam - does chaining requests improve speed, or just make it easier coding?

PR #271, opened 2021-05-24
Added "locations" field to communities.list route. Working well.

frimpongopoku:
`communities.list` did not include locations to communities, so I added that field in this PR, and it works with the latest frontend PR.


BradHN1 (commented):
Need to pass back "location", not "locations".

BradHN1 (approved):
Perfect!

PR #276, opened 2021-06-04
Included vendor logo in vendor info obj, so that testimonial vendor c…

frimpongopoku:
Included vendor logo in vendor info obj, so that testimonial vendor can have the vendor logo


BradHN1 (approved):
Good

PR #279, opened 2021-06-07
Implemented profile picture

isha-san:


src/_main_/settings.py
```diff
@@ -24,7 +24,7 @@
 # ********  LOAD CONFIG DATA ***********#
 IS_PROD = False
 IS_CANARY = False
-IS_LOCAL = False
+IS_LOCAL = True
```
BradHN1, line 27: try to avoid committing this or DEBUG=True

src/_main_/settings.py
```diff
@@ -236,7 +237,9 @@
 # Static files (CSS, JavaScript, Images)
 STATIC_URL = '/static/'
 MEDIA_URL = '/media/'
+MEDIA_ROOT = os.path.join(BASE_DIR, 'media')
 
+STATIC_ROOT = os.path.join(BASE_DIR, 'static')
```
BradHN1, line 242: Do these definitions of MEDIA_ROOT and STATIC_ROOT change where those files get deployed and could the change screw things up?

isha-san: Based on the resources online that I looked at, the answer is no.

src/_main_/settings.py
isha-san: Noted

src/api/services/auth.py
```diff
@@ -84,8 +84,10 @@ def whoami(self, context: Context):
 
       if user and context.is_admin_site and not(user.is_super_admin or user.is_community_admin):
         raise PermissionError
-
-      return serialize(user, full=True), None
+      #print(user)
+      s = serialize(user, full=True)
+      print(s)
+      return s, None
```
BradHN1, line 90: Usually don't commit the print statements (except those which are helpful going forward)

src/api/store/userprofile.py
```diff
@@ -205,20 +205,21 @@ def create_user(self, context: Context, args) -> (dict, MassEnergizeAPIError):
...
+      print(user)
       if not user:
         new_user: UserProfile = UserProfile.objects.create(
           full_name = args.get('full_name'), 
           preferred_name = args.get('preferred_name', None), 
           email = args.get('email'), 
           is_vendor = args.get('is_vendor', False), 
-          accepts_terms_and_conditions = args.pop('accepts_terms_and_conditions', False)
+          accepts_terms_and_conditions = args.pop('accepts_terms_and_conditions', False), 
+          color = args.get('color')
```
BradHN1, line 218: This will blow up if 'color' not specified, which can be the case.
For args which are required, check out the self.validator.expect('argname', argtype, required=True) in the handler routine.

src/database/migrations/0093_auto_20210607_0945.py
```diff
@@ -0,0 +1,22 @@
...
+        migrations.RemoveField(
+            model_name='userprofile',
+            name='auto_icon',
+        ),
+        migrations.AddField(
+            model_name='userprofile',
+            name='color',
+            field=models.CharField(default='#000000', max_length=7),
+        ),
+    ]
```
BradHN1, line 22: I would consolidate the three migration files, by deleting them and regenerating

src/database/models.py
```diff
@@ -519,6 +519,7 @@ class UserProfile(models.Model):
   created_at = models.DateTimeField(auto_now_add=True)
   updated_at = models.DateTimeField(auto_now=True)
   is_deleted = models.BooleanField(default=False, blank=True)
+  color = models.CharField(default="#000000", max_length = 7)
```
BradHN1, line 522: I'm thinking that it would make sense to keep the color pref in a JSON field user_prefs.  We'll be adding more prefs for communication and viewing site stuff.

src/database/views.py
```diff
@@ -1,3 +1,7 @@
 from django.shortcuts import render
 
 # Create your views here.
+
+class UserAvatarUpload(APIView):
```
BradHN1, line 5: I can't find where this is used

src/api/handlers/userprofile.py
```diff
@@ -117,8 +117,8 @@ def remove_user_action(self, request):
   def update(self, request):
     context: Context = request.context
     args: dict = context.args
-    args = rename_field(args,'id','user_id')
-    user_id = args.pop('user_id', None)
+    user_id = args.get('user_id', None)
```
BradHN1, line 120: A small bug.  For this and other routes one can pass in 'id' or 'user_id'.

src/api/handlers/userprofile.py
```diff
@@ -61,6 +60,7 @@ def create(self, request):
       .expect("full_name", str, is_required=True)
       .expect("preferred_name", str, is_required=True)
       .expect("is_vendor", bool, is_required=True)
+      .expect('color', str, is_required=True)
```
BradHN1, line 63: Color should not be a required parameter.  This will break one other app which uses this route.

src/api/store/userprofile.py
```diff
@@ -205,20 +205,21 @@ def create_user(self, context: Context, args) -> (dict, MassEnergizeAPIError):
...
+      print(user)
       if not user:
         new_user: UserProfile = UserProfile.objects.create(
           full_name = args.get('full_name'), 
           preferred_name = args.get('preferred_name', None), 
           email = args.get('email'), 
           is_vendor = args.get('is_vendor', False), 
-          accepts_terms_and_conditions = args.pop('accepts_terms_and_conditions', False)
+          accepts_terms_and_conditions = args.pop('accepts_terms_and_conditions', False), 
+          preferences = {'color': args.get('color')}
```
BradHN1, line 218: only save the color if specified as a parameter.
In the front end, if the color isn't specified, and the profile  picture isn't uploaded, let's prompt them (once) to choose a color or invite to upload a profile pic

src/database/models.py
```diff
@@ -495,14 +495,14 @@ class UserProfile(models.Model):
   created_at: DateTime
     The date and time of the last time any updates were made to the information
     about this goal
-
+  color:  models.CharField(default="#000000", max_length = 7)
```
BradHN1, line 498: comment- I removed

PR #285, opened 2021-07-02
Merge Isha's profile pic changes and my clean-ups

BradHN1:


isha-san (commented):
Should we change the settings.py variables back to match that it will be deployed?

isha-san (approved):

PR #286, opened 2021-07-02
Import contacts

isha-san:


src/api/handlers/auth.py
```diff
@@ -28,6 +28,7 @@ def registerRoutes(self):
   
   def login(self, request): 
     context: Context = request.context
+    print(context)
```
BradHN1, line 31: don't commit prints

isha-san: ah yes i will fix this now sorry

BradHN1 (approved):
Ready to test

PR #288, opened 2021-07-10
flipped `null` to `True` on new `Preferences` field -- UserProfile.

frimpongopoku:


Opoku-Agyemang (approved):

BradHN1 (commented):
There are three migration files included, and one line change in the models.py file.  This doesn't make sense to me; looks like there was an error in the DonatePageSettings temporarily which wasn't consistent with the database?

PR #289, opened 2021-07-17
Recurring events

isha-san:
Allow CAdmins to create recurring events. Docs here: https://docs.google.com/document/d/1PF6AAtswPTY7CMkDAf-DKUOcBaU1J9e-6wYO5aI6h8g/edit?usp=sharing


src/api/handlers/userprofile.py
```diff
@@ -233,4 +233,6 @@ def list_events(self, request):
     user_info, err = self.service.list_events_for_user(context, args)
     if err:
       return MassenergizeResponse(error=str(err), status=err.status)
+    print('ehllo')
```
BradHN1, line 236: prints

src/website/tasks.py
```diff
@@ -0,0 +1,10 @@
+# celery tasks specific to api app
```
BradHN1, line 1: @isha-san - what does this file do?  Did you mean to save it?

src/api/store/event.py
```diff
@@ -54,70 +61,175 @@ def copy_event(self, context: Context, args) -> (dict, MassEnergizeAPIError):
...
 
 
   def create_event(self, context: Context, args) -> (dict, MassEnergizeAPIError):
+    
     try:
       image = args.pop('image', None)
       tags = args.pop('tags', [])
       community = args.pop("community_id", None)
-
+      is_recurring  = args.pop('is_recurring', False)
```
BradHN1, line 125: looks like if "is_recurring" not passed in, and interpreted as False (not "false" or "") that recurring will be True

src/api/store/event.py
```diff
@@ -54,70 +61,175 @@ def copy_event(self, context: Context, args) -> (dict, MassEnergizeAPIError):
...
+      if is_recurring == "false" or is_recurring == "":
+        recurring = False
+      else: recurring = True
+      recurring_type = args.pop('recurring_type', None)
+      separation_count = args.pop('separation_count', None)
```
BradHN1, line 130: I think this will blow up if separation_count not passed.  int(None) gives an exception

BradHN1 (approved):
I'll make a couple fixes as noted

PR #291, opened 2021-07-25
Fixes for creating and updating events

BradHN1:


isha-san (approved):

src/api/store/event.py
```diff
@@ -2,14 +2,62 @@
...
 from _main_.utils.context import Context
 from sentry_sdk import capture_message
 from .utils import get_user_or_die
+#from _main_.utils.common import parse_date
 import datetime
 from datetime import timedelta
 import calendar
 import pytz
 
+def _check_recurring_date(start_date_and_time, end_date_and_time, day_of_week, week_of_month):
```
Opoku-Agyemang, line 15: this is the kind of thing i would loooooove to see some tests for.  What do you think about adding some good ol tests to capture this?

Opoku-Agyemang (changes requested):

src/api/store/event.py
BradHN1: Isha added some tests

BradHN1:
@Opoku-Agyemang - please review, and deploy to dev if you want.  Dev currently broken because of a model inconsistency (fixed in the last pushed change)

PR #292, opened 2021-07-27
Import contacts

isha-san:


src/api/handlers/userprofile.py
```diff
@@ -267,17 +268,29 @@ def complete_imported_user(self, request):
...
-    # so we copy it as a csv temporarily to the disk
-    temporarylocation="testout.csv"
-    with open(temporarylocation, 'wb') as out:
-      var = csv_ref.read()
-      out.write(var)
-    info, err = self.service.handle_csv(context, args, temporarylocation)
+    try:
+    
+      # find the community within the team that the 
+      csv_ref = args['csv'].file
```
Opoku-Agyemang, line 274: nit: a bit unsafe.  
Either asset that csv is inside args or use `args.get('csv')` and maje sure its not None before you  try to get file from it.

src/api/handlers/userprofile.py
```diff
@@ -267,17 +268,29 @@ def complete_imported_user(self, request):
...
+      temporarylocation="testout.csv"
+      with open(temporarylocation, 'wb') as out:
+        var = csv_ref.read()
+        out.write(var)
+      # filecontents is a list of dictionaries, with each list item representing a row in the CSV
+      filecontents = []
+      with open(temporarylocation, "r") as f:
+          reader = csv.DictReader(f, delimiter=",")
+          for row in reader:
+            filecontents.append(row)
```
Opoku-Agyemang, line 286: this logic seems like it would be useful as a helper function.  can you make a helper function in utils/common.py and define this there?

src/api/services/userprofile.py
```diff
@@ -152,8 +153,26 @@ def add_action_completed(self, context, args) -> (dict, MassEnergizeAPIError):
...
+    email_field = args.get('email_field', None)
+    invalid_emails = []
+    for csv_row in filecontents:
+      column_list = list(csv_row.keys())
+      try:
+        # prevents the first row (headers) from being read in as a user
+        if csv_row[first_name_field] == column_list[0]:
+          continue
+        # verify correctness of email address
+        regex = '^[a-z0-9]+[\._]?[a-z0-9]+[@]\w+[.]\w{2,3}$'
```
Opoku-Agyemang, line 167: again a helper function called `is_valid_email` would be great so that other functions can call that same function in other places.

src/api/services/userprofile.py
```diff
@@ -152,8 +153,26 @@ def add_action_completed(self, context, args) -> (dict, MassEnergizeAPIError):
...
+        if(re.search(regex,csv_row[email_field])):   
+          info, err = self.store.import_from_csv(context, args, csv_row)
+        else:
+          if filecontents.index(csv_row) != 0:
+            invalid_emails.append(csv_row + 1) 
+      except Exception as e:
+        print(str(e))
```
Opoku-Agyemang, line 174: nit: remove print

src/api/services/userprofile.py
```diff
@@ -152,8 +153,26 @@ def add_action_completed(self, context, args) -> (dict, MassEnergizeAPIError):
...
+        return None, CustomMassenergizeError(str(e))
+    if err:
+      return None, err
+    return {'invalidEmails': invalid_emails}, None
```
Opoku-Agyemang, line 178: always leave an empty line at the bottom

src/api/store/userprofile.py
```diff
@@ -575,62 +575,42 @@ def handle_csv(self, context: Context, args, temporarylocation) -> (dict, MassEn
...
+      message = cadmin.full_name + " invited you to join the following MassEnergize Community: " + registered_community.name + "\n"
+      mess = args.get('message', None)
+      if mess and mess != "":
+        message += "They have included a message for you here:\n"
+        message += mess
+      if team:
+        message += "You have been assigned to the following team: " + team.name + "\n"
+      link = "massenergize.org/" + str(registered_community.subdomain) + "/signup"
+      message += "Use the following link to join " + registered_community.name + ": " + link
+      send_massenergize_email(subject= cadmin.full_name + " invited you to join a MassEnergize Community", msg=message, to=new_user.email)
```
Opoku-Agyemang, line 611: we should really not be doing any  business logic things inside the store file.  We've done very poorly in the past and i think we have a lot of tech debt in moving all those logic from the store files into the service files.  

In this scenario, call the store to just save data or fetch data and that's it.  Dont send emails in store, don't do multiple things in the same function too.  Functions should usually do one thing and only one thing.

Opoku-Agyemang (changes requested):
Great progress!
Left a few comments for you

PR #294, opened 2021-08-04
Fix pageSetting images

BradHN1:
Model supports multiple images, though code currently supports single.
No model changes required.

### DIFF
--- a/src/api/store/page_settings.py
+++ b/src/api/store/page_settings.py
@@ -1,4 +1,4 @@
-from database.models import UserProfile, Community
+from database.models import UserProfile, Community, Media
 from _main_.utils.massenergize_errors import MassEnergizeAPIError, InvalidResourceError, ServerError, CustomMassenergizeError
 from _main_.utils.massenergize_response import MassenergizeResponse
 from _main_.utils.context import Context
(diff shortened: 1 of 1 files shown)


src/api/store/page_settings.py
```diff
@@ -61,10 +61,15 @@ def update_page_setting(self, page_setting_id, args) -> (dict, MassEnergizeAPIEr
     page_setting = page_setting.first()
 
     if image:
-      media = Media.objects.create(name="Page-Image", file=image)
-      page_setting.image = media
+      # Page settings models support multiple images, but we currently allow just one.
+      if not image == "None":
```
sopokuagyemang, line 65: i think we should rather make it clearer by doing 
if image == "None" then we clear the thing else: do something else

sopokuagyemang (changes requested):
nit comment

src/api/store/page_settings.py
```diff
@@ -61,10 +61,15 @@ def update_page_setting(self, page_setting_id, args) -> (dict, MassEnergizeAPIEr
...
+        media = Media.objects.create(name="Page-Image", file=image)
+        page_setting.images.clear()
+        page_setting.images.add(media)
+      else:
+        page_setting.images.clear()
+        #page_setting.images = None
```
sopokuagyemang, line 71: remove if not necessary or add a TODO comment on why u r leaving it there

PR #309, opened 2021-09-03
Sam seo patch 2

Opoku-Agyemang:
In this PR i  do the following:
1. cleanup the webpages 
2. Introduce subdomain reservation since we will be using it for routing communities properly
3. subdomain backfills
4. subdomain reserved words
3. sitemap updates
4. make all images stored public since we are not yet storing any sensitive files.  This should make it easier for us to cache and do other interesting work in the future


BradHN1:
Can’t test this now. But make sure api urls work after change to urls.py

massenergize-admin (approved):

Issue #322, opened 2021-09-25
CAN: create action: Carbon calculator dropdown boxes empty


kaatvds:
For both Super and Community Admin the only option is "select"...

Issue #321, opened 2021-09-24
Communities.update returns error


BradHN1:
the routine reserve_subdomain tries to create a Subdomain, even if it already exists.  THis gives a duplicate key error.

Issue #320, opened 2021-09-24
Communities.listForCommunityAdmin should return all communities if a super-admin


BradHN1:
If a user is both a cadmin and a superadmin, the all communities list only shows communities that person is a cadmin for.  This is from a logic flaw where it checks first if the user is a cadmin, then whether the user is a sadmin.

PR #323, opened 2021-09-27
Issues 320, 321, 322

BradHN1:
#320 320 communities.listForCommuityAdmin returns all for a super admin (even if also a cadmin);
#321 Fix error from reserve_subdomain
#322 Create action - carbon calc box empty

### DIFF
--- a/src/api/store/community.py
+++ b/src/api/store/community.py
@@ -441,8 +441,8 @@ class CommunityStore:
         self, context: Context, args
     ) -> Tuple[dict, MassEnergizeAPIError]:
         try:
-            subdomain = args.get("subdomain")
-            community_id = args.get("id")
+            subdomain = args.get("subdomain", None)
+            community_id = args.get("id", None)
 
             if not community_id and not subdomain:
                 return None, CustomMassenergizeError(
@@ -765,7 +765,7 @@ class CommunityStore:
                 community.favicon = cFavicon
                 community.save()
 
-            if subdomain:
+            if subdomain and not Subdomain.objects.filter(name=subdomain):
                 reserve_subdomain(subdomain, community)
             
             return community, None
@@ -800,12 +800,12 @@ class CommunityStore:
         try:
             # if not context.user_is_community_admin and not context.user_is_community_admin:
             #   return None, CustomMassenergizeError("You are not a super admin or community admin")
-            if context.user_is_community_admin:
+            if context.user_is_super_admin:
+                return self.list_communities_for_super_admin(context)
+            elif context.user_is_community_admin:
                 user = UserProfile.objects.get(pk=context.user_id)
                 admin_groups = user.communityadmingroup_set.all()
                 return [a.community for a in admin_groups], None
-            elif context.user_is_super_admin:
-                return self.list_communities_for_super_admin(context)
             else:
                 return [], None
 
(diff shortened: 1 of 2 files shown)


src/api/store/community.py
```diff
@@ -853,13 +853,23 @@ def can_use_this_subdomain(subdomain: str, community: Community=None) -> bool:
     return False
 
 def reserve_subdomain(subdomain: str, community: Community=None):
+    # first check that we can use this domain
```
BradHN1, line 856: why don't we first check whether the Subdomain exists for the community and already is in use, and if so return without doing anything.  This will be the case 99% of the time.

src/api/store/community.py
```diff
@@ -853,13 +853,23 @@ def can_use_this_subdomain(subdomain: str, community: Community=None) -> bool:
...
+        return False
+
+    # if we are here then the subdomain is available to be used by this community
+    # now let's make sure it is all lower case
     subdomain = subdomain.lower()
 
-    if subdomain in RESERVED_SUBDOMAIN_LIST:
-        return False
-    # mark the old subdomains for this community to unused
+    # mark the old subdomains for this community to un-used
```
BradHN1, line 864: Problem with this (if you don't return earlier) is you mark the Subdomain the community is using as 'not used'

src/api/store/community.py
```diff
@@ -853,13 +853,23 @@ def can_use_this_subdomain(subdomain: str, community: Community=None) -> bool:
...
-    new_subdomain.save()
+    # let's do a search for if subdomain is already created for this community
+    subdomain_search = Subdomain.objects.filter(community=community, subdomain__iexact=subdomain)
+    if not subdomain_search.exists():
+        # if subdomain is not really created then create a new one
+        new_subdomain = Subdomain(name=subdomain, community=community, in_use=True)
+        new_subdomain.save()
+    else:
+        # if we are here then the subdomain already exists for this community so let's flip it to in-use
+        Subdomain.objects.filter(community=community, in_use=False)
```
BradHN1: maybe there is another line here I can't see, but this doesn't flip it to in-use.  Need to do the .update(in_use=True)

src/api/store/community.py
Opoku-Agyemang: we set it on line 875

src/api/store/community.py
Opoku-Agyemang: It is actually the same thing that is implemented.  I can make it return early but the outcome is the same.  I just wanted to reuse the function above to handle that case.

BradHN1: It doesn't really do the same thing.  Remember most of the time the cadmin will be changing something else, very rarely will the subdomain change.  So in most cases, can_use_this_subdomain will return true, then you'll mark it as unused, then you'll search for it, find it, and mark it as used.  Why not just check it first and do nothing if it exists and is used?

Opoku-Agyemang: the outcome is the same.  but i am happy to refactor to fail early

src/api/store/community.py
```diff
@@ -861,7 +861,7 @@ def reserve_subdomain(subdomain: str, community: Community=None):
     subdomain = subdomain.lower()
 
     # if subdomain is in use for this community just return right away
-    if Subdomain.objects.filter(community=community, subdomain__iexact=subdomain, in_use=True).update(in_use=False):
+    if Subdomain.objects.filter(community=community, subdomain__iexact=subdomain, in_use=True).update(in_use=True):
```
BradHN1: Better, but you don't need the .update(in_use=True)

sopokuagyemang (approved):

Issue #325, opened 2021-09-29
API Community Data Changes


BradHN1:
Two parts:

1, set community goals and initial values with the graphs.data.update route, together with setting the category input data

2. Eliminate the 'initial_number_of_actions' so donut graphs are consistent with the actions by category graph.

PR #326, opened 2021-09-30
API Data Changes

BradHN1:
Issue #325 - for Admin portal to consolidate input data into one form, set community goals and initial values in graphs.data.update route


Opoku-Agyemang (approved):
you can remove the commented code.

Issue #367, opened 2021-11-04
Super Admin downloads broken - All Users & All Communities


apowelka:
When logged in as a Super Admin, two data downloads do not work but have different errors. The third, All Actions, is working just fine.

page: https://admin.massenergize.org/
![image](https://user-images.githubusercontent.com/60662056/140396386-6c0935c3-f9bf-4763-b4e0-0ac45d2a3107.png)

ALL USERS - doesn't complete, just keeps spinning circle.
ALL COMMUNITIES - get an html download, not csv

BradHN1:
The all communities download is easy to fix.
The ALL USERS download is already an issue 259, so will be fixed in that ticket.

Issue #357, opened 2021-10-12
PROD: messages to admins NOT forwarded in email


kaatvds:
creating new ticket: This is Medfield's bug. I tested it and I also don't get the message.

BradHN1:
Change based on complaint from Medfield: instead of sending the e-mail to all community admins for a site, which could be several people, send it to the primary contact for the site (community.owner_name, community.owner_email). One person should be responsible for either responding or getting someone else to respond.

PR #373, opened 2021-11-11
Fix team users download and add some unit tests

BradHN1:
#367 User downloads broken

### DIFF
--- a/src/_main_/settings.py
+++ b/src/_main_/settings.py
@@ -18,6 +18,12 @@ from pathlib import Path  # python3 only
 import sentry_sdk
 from sentry_sdk.integrations.django import DjangoIntegration
 
+
+DJANGO_ENV = os.environ.get("DJANGO_ENV","remote")
+DEV_SERVER = False
+if DJANGO_ENV == "local":
+    DEV_SERVER = True
+
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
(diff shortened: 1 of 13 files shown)


BradHN1:
Fix two other issues in this PR:
#357 - contact cadmin - send e-mail just to primary contact not all cadmins
#374 - problem running locally

src/_main_/settings.py
```diff
@@ -18,6 +18,12 @@
 import sentry_sdk
 from sentry_sdk.integrations.django import DjangoIntegration
 
+
+DJANGO_ENV = os.environ.get("DJANGO_ENV","remote")
+DEV_SERVER = False
```
Opoku-Agyemang, line 23: can we use a name like `RUN_SERVER_LOCALLY`
Also can we add comments to describe what each of these stand for?

src/_main_/settings.py
```diff
@@ -63,6 +70,7 @@
     'massenergize-canary-api.us-east-2.elasticbeanstalk.com',
     'massenergize.test',
     'massenergize.test:3000',
+    'communities.massenergize.test'
```
Opoku-Agyemang, line 73: massenergize.test covers this OR you could also do *.massenergize.test

Opoku-Agyemang (changes requested):
I think we are mixing several things in one PR which makes it hard to review and isolate things.  Happy to +1 but its risky

src/_main_/settings.py
BradHN1: ok

src/_main_/settings.py
BradHN1: Yes, done

BradHN1:
You're right, my bad.
Most of them are pretty simple fixes.

Opoku-Agyemang (approved):

PR #376, opened 2021-11-16
Pongo media libary implementation

frimpongopoku:
Media library core routes setup in this PR

**Also ignore the git change count, its just cos I formatted the** `models.py` 🤣 🤣 🤣  

The only real changes I made are around `UserMediaProfile`, `Actions` ,`Events` `Testimonials` and `Media`


src/database/models.py
```diff
@@ -1,2613 +1,2917 @@
...
 import uuid
 
-CHOICES = json_loader('./database/raw_data/other/databaseFieldChoices.json')
-ZIP_CODE_AND_STATES = json_loader('./database/raw_data/other/states.json')
+CHOICES = json_loader("./database/raw_data/other/databaseFieldChoices.json")
+ZIP_CODE_AND_STATES = json_loader("./database/raw_data/other/states.json")
 
 
 class Location(models.Model):
-  """
```
Opoku-Agyemang, line 13: i think u may have formatted this file with ur editor.  Can u undo?

src/api/store/media_library.py
```diff
@@ -0,0 +1,123 @@
...
+            )
+        return images, None
+
+    def generateQueryWithScope(self, scope, com_id):
+        queries = {
+            "actions": Media.objects.filter(actions__community__id=com_id),
+            "events": Media.objects.filter(events__community__id=com_id),
+            "testimonials": Media.objects.filter(testimonials__community__id=com_id),
+            "uploads": Media.objects.filter(user_upload__community__id=com_id),
+            "default": Media.objects.filter(
```
Opoku-Agyemang, line 51: why do the same job twice?  the queries are already expensive as they are.

You may want to do 
a = ...
b = ....
c = ....
d = ...

default = a|b|c|d

You get the idea?

src/api/store/media_library.py
```diff
@@ -0,0 +1,123 @@
...
+                | Q(testimonials__community__id=com_id)
+                | Q(user_upload__community__id=com_id)
+            ),
+        }
+        query = queries.get(scope)
+        if not query:
+            query = queries.get("default")
+        return query
+
+    def search(self, args):
```
Opoku-Agyemang, line 63: always use .get() as you could run into key error

Opoku-Agyemang (changes requested):

src/database/models.py
frimpongopoku: yes, that is what I explained earlier, and No, it is too far ahead to undo. Also, is there a reason we are choosing not to format that file? Can you come up with a default indentation that everyone can use, so that this isnt a problem anymore?

src/api/store/media_library.py
frimpongopoku: I read that when you write queries like this, django actually doesnt fetch anything, until you use it. When you write it like this, it actually just builds only the queries. 🤔

frimpongopoku: Also I dont think `a|b...` fits in this particular scenario, cos we want to only be finding images related to "actions" when we specify the scope as "actions" 🤔

.gitignore
```diff
@@ -17,3 +17,6 @@ src/Dockerrun.aws.json
 .idea
 src/api/tests/TestCases.json
 src/api/store/ZIPCODE_FIXES.json
+**/**/migrations
```
BradHN1, line 20: migrations ar an important part of repository.  Shouldn't be in .gitignore

.gitignore
```diff
@@ -17,3 +17,6 @@ src/Dockerrun.aws.json
...
+**/**/settings.py
+src/_main_/settings.py
```
BradHN1, line 22: settings.py is an important part of the repository.  Not to be ignored.  Just don't commit it if you don't want your changes.

frimpongopoku:
I added migrations in .gitignore because they are always causing trouble.
🤣🤣
So its better if they are regenerated when needed isnt it?


For settings as well, nothing much changes overthere... So i just thought
it would be okay to not add to commits unless any proper changes occur....
Dont you agree?

src/database/models.py
BradHN1: Agree with Sam, makes it very hard to see what the changes are when the entire file changes.  
That said, if we decide to do a standard formatting (that we agree to), let's do that as a separate PR

src/api/handlers/auth.py
```diff
@@ -1,5 +1,6 @@
 """Handler file for all routes pertaining to auths"""
 
+from _main_.settings import IS_LOCAL
```
BradHN1, line 3: Can you leave this out; contradicts what was just merged the other day.

BradHN1:
I think when you add something to .gitignore, it means I can’t commit them either.  They need to be part of the repository.
Better that you just don’t commit files (including settings.py or config.json) if you are just modifying for local testing.

src/_main_/utils/utils.py
```diff
@@ -3,72 +3,92 @@
...
+def strip_website(url: str) -> str:
+    if not url:
+        return None
+    url = url.replace("http://", "")
+    url = url.replace("https://", "")
+    url = url.split("/")[0]  # deal with trailing spaces
+    return url.strip()
+
+
+class Console:
```
BradHN1, line 85: Not sure what this is about?  Just for local testing?

frimpongopoku: Yes, it wont affect the product...  🤣 😂😂😂😂 just makes it easy to find the things I print.... the terminal logs too many things.

 Also I think I will just close this PR in the mean time... I am still working on it and I forgot you would be notified of the tiny changes 🤣 🤣 ....

BradHN1:
Why don’t you convert it to “Draft” for now.  I don’t mind the notifications, it is good to see what is up.

PR #386, opened 2021-11-24
Anonymous profiles

Davidjwkinghorn:


src/api/handlers/deviceprofile.py
```diff
@@ -113,6 +115,75 @@ def log_device(self, request):
...
+    args: dict = context.args
+
+    self.validator.expect("metric", str, is_required=True)
+    self.validator.expect("community_id", str)
+    args, err = self.validator.verify(args)
+
+    if err:
+      return err
+
+    metric = args["metric"] if "metric" in args else None
```
BradHN1, line 131: you can do this with args.get("metric", None)

BradHN1 (approved):
Looks good

src/api/handlers/deviceprofile.py
```diff
@@ -94,7 +96,7 @@ def log_device(self, request):
     if err:
       return err
 
-    self.__get_client_meta(request, args)
+    # self.__get_client_meta(request, args)
```
Opoku-Agyemang, line 99: remove all commented out code if not needed.

If you wish to leave any in there, leave a comment on why you prefer to keep the deadcode

Opoku-Agyemang (changes requested):

src/api/handlers/deviceprofile.py
```diff
@@ -113,6 +115,75 @@ def log_device(self, request):
...
+        
+    return MassenergizeResponse(data=metric)
+  
+  @admins_only
+  def community_metrics(self, request):
+    context: Context = request.context
+    args: dict = context.args
+
+    self.validator.expect("metric", str, is_required=True)
+    self.validator.expect("community_id", str)
```
BradHN1, line 162: This route looks like it would pass junk back if community_id not specified.  Should it be required?

BradHN1: and if the metric is something different than the three choices, should it return an error?

BradHN1 (approved):
Looks good - work in progress

src/api/handlers/deviceprofile.py
Davidjwkinghorn: I had meant to get round to that. I must have missed a few of those. I'll go back and check.

BradHN1 (approved):
Very good

PR #394, opened 2021-12-08
Backend Changes for Issue 400 (Front end)

dim505:


src/api/handlers/testimonial.py
```diff
@@ -12,149 +12,156 @@
...
-
-  def info(self, request):
-    context: Context = request.context
-    args: dict = context.args
-
-    self.validator.expect("id", int, is_required=True)
-    self.validator.rename("testimonial_id", "id")
-    args, err = self.validator.verify(args)
-
-    if err:
```
Opoku-Agyemang, line 43: It seems you're reformatting the whole file by indenting with two more spaces.  Could you revert?  It makes it hard to nit pick what your real changes are.

Opoku-Agyemang (changes requested):
Also, please add some context and description to your PR.  I don't know what its about.

PR #418, opened 2022-01-11
417 Minor fixes for action dates

BradHN1:
A few fixes relating to dates and UserActionRels

### DIFF
--- a/src/_main_/utils/common.py
+++ b/src/_main_/utils/common.py
@@ -81,7 +81,11 @@ def parse_int(b):
 
 def parse_date(d):
   try:
-    return pytz.utc.localize(datetime.strptime(d, '%Y-%m-%d %H:%M'))
+    if len(d) == 10:
+      return pytz.utc.localize(datetime.strptime(d, '%Y-%m-%d'))
+    else:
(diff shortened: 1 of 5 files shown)


src/database/models.py
```diff
@@ -797,6 +799,8 @@ def update_visit_log(self, date_time):
                 new = []
                 for day in old.keys():
                     old_format = "%d/%m/%Y"
+                    if len(day) < 10:
```
Opoku-Agyemang, line 802: this feels very weird

Opoku-Agyemang (approved):

src/database/models.py
BradHN1: There were a couple cases in the dev DB where the date got entered as a 2 digit year, this cleans that up.

src/_main_/utils/common.py
```diff
@@ -81,7 +81,11 @@ def parse_int(b):
 
 def parse_date(d):
   try:
-    return pytz.utc.localize(datetime.strptime(d, '%Y-%m-%d %H:%M'))
+    if len(d) == 10:
```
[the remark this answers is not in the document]
BradHN1: I'm not a fan either, but before the code would barf on any date without exactly 16 characters, and change the date to be the current date.

PR #429, opened 2022-01-26
A couple minor bugs fixed, plus the menus are right

BradHN1:


frimpongopoku (approved):

src/api/handlers/deviceprofile.py
```diff
@@ -57,24 +58,26 @@ def log_device(self, request):
...
-      ip_address = self.ipLocator.getIP(request)
-      if not ip_address or ip_address in ['127.0.0.1']:
-        ip_address = "38.242.8.93" # For testing
+    if RUN_SERVER_LOCALLY or context.is_admin_site:
+      return MassenergizeResponse(data=None)
 
-      args["ip_address"] = ip_address  
+    ip_address = self.ipLocator.getIP(request)
+    if not ip_address or ip_address in ['127.0.0.1']:
+      ip_address = "38.242.8.93" # For testing
```
Opoku-Agyemang, line 66: I'm not sure i understand why we are setting this ip address.  Why not keep it as the localhost ip?

Opoku-Agyemang (changes requested):

src/api/handlers/deviceprofile.py
BradHN1: That happens to be my IP address.  The localhost IP gives an error result when used with the geolocator.  Since it is only for local testing it doesn't really matter.

Davidjwkinghorn (approved):

PR #439, opened 2022-02-22
Updates to DeviceProfile logging

BradHN1:
add community, don't create new ID if device exists

note questions in comments

### DIFF
--- a/src/api/store/deviceprofile.py
+++ b/src/api/store/deviceprofile.py
@@ -1,17 +1,10 @@
-from calendar import month
-from django.http import HttpRequest
 from datetime import datetime
-from api.handlers.userprofile import UserHandler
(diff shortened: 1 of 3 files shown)


src/api/store/deviceprofile.py
```diff
@@ -51,6 +44,7 @@ def create_device(self, context: Context, args, save=True) -> Tuple[dict, MassEn
     try:
       new_device: DeviceProfile = DeviceProfile.objects.create(**args)
       
+      # question: is this needed?  those values not already part of new_device through create with the args?
```
Davidjwkinghorn, line 47: The __device_attr_handler() method is just a way of reducing duplicate code. It's technically not needed, but having lots of duplicate code is considered a bad practice.

If the checks in this method aren't needed we could remove it. When I wrote it I was mimicking another api store.

PR #446, opened 2022-02-24
Issues 444, 445

BradHN1:
API routes for user submitted actions, events, vendors
Fix issues with copying actions, events, vendors on admin portal


src/api/store/vendor.py
```diff
@@ -6,9 +6,24 @@
...
+  # Community will be None at first
+  new_title = get_new_title(None, old_title)
+  version = 0
+  while version<1000:
+    suffix = "-Copy%d" % version
+    newlen = min(len(new_title), SHORT_STR_LEN - len(suffix))
+    title = new_title[0:newlen] + suffix
+    if not Vendor.objects.filter(name=title).first():
+      return title
+    version += 1
```
sopokuagyemang, line 25: Could we explore other options here?  
In the worst case this function seems like it will query the DB 1000 times just to create a title.

Also i am not sure of the goal exactly

src/api/store/vendor.py
```diff
@@ -196,13 +223,23 @@ def copy_vendor(self, vendor_id) -> Tuple[Vendor, MassEnergizeAPIError]:
...
-      return vendor, None
+
+      old_tags = vendor.tags.all()
+      new_vendor = vendor
+
+      new_vendor.pk = None
+      new_vendor.is_published = False
+      new_vendor.is_verified = False
+
+      new_name = _copy_title(vendor.name)
```
sopokuagyemang, line 234: i see that you got rid of the random number generator.  I'm not sure the replacement is the best path forward here either.

src/website/templates/action_submitted_email.html
```diff
@@ -0,0 +1,203 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
```
sopokuagyemang, line 1: i'm guessing these emails are all in the postmark as well?

sopokuagyemang: If so in our email helper function can we do it such that while in production, the emails are sent from postmark and while in local production mode, we send via the django email client?

src/api/store/event.py
```diff
@@ -6,9 +6,25 @@
...
+  # Community will be None at first
+  new_title = get_new_title(None, old_title)
+  version = 0
+  while version<1000:
+    suffix = "-Copy%d" % version
+    newlen = min(len(new_title), SHORT_STR_LEN - len(suffix))
+    title = new_title[0:newlen] + suffix
+    if not Event.objects.filter(name=title).first():
+      return title
+    version += 1
```
sopokuagyemang, line 26: Same comments as here as above.
Also, something like this should always go with tests to accompany.

sopokuagyemang (changes requested):

src/api/store/vendor.py
BradHN1: The goal is to pick the first available title that doesn't duplicate one already used.  I don't like the random number solution, because from the community admin perspective it looks like a hack.  The usual case is that a community admin copies an action/event/vendor and the first try ("-Copy0") works, then they edit the action/event/vendor to be for their community and change the name to what they want.  In a rare case, someone might Copy something several times without changing the name, and we'd have a few copies.  Not 1000, and we can have the chron job go through monthly and delete extraneous copies of actions/events/vendors that aren't connected with any communities or published.
Certainly could be better solutions, but we don't want to go back to the random numbers.

src/website/templates/action_submitted_email.html
BradHN1: David set up 3 postmark servers, for production, development and testing.  I think this will work well, so we can see in any case what message went out.

src/api/store/event.py
BradHN1: I didn't add tests, but the unit testing does check the copy routes.

src/api/store/vendor.py
sopokuagyemang: Hmmm i don't think we want this either.
First the query itself is filtering by a column that is not indexed. 
Second, we will do this in a loop that could possibly go a 1000 times.
Considering these are all network calls and will incur overheads, i don't think we should go with this approach either.  I'd say until we have a better solution that random numbers, let's not do anything to compromise on performance.

src/api/store/event.py
sopokuagyemang: also it seems the github tests action is failing.  The pillow version in requirements.txt is not correct i suppose.  Could you update it so the tests can run ?

src/api/store/vendor.py
```diff
@@ -196,13 +213,28 @@ def copy_vendor(self, vendor_id) -> Tuple[Vendor, MassEnergizeAPIError]:
...
+
+      new_vendor.pk = None
+      new_vendor.is_published = False
+      new_vendor.is_verified = False
+
+      # the copy will have "-Copy" appended to the name; if that already exists, delete it first
+      new_name = vendor.name + "-Copy"
+      check = Vendor.objects.filter(name=new_name).first()
+      if check:
+        check.delete()
```
Opoku-Agyemang, line 228: Hmmmm just checked it out.  This approach also seems a bit dangerous.  The vendor which is the previous copy in this case could be linked to so many other models and instances.  Ideally, instead of deleting an old copy and basically creating it as a new one, it may be better to rather
* Copy over the latest updates into the copy that already exists and return the old copy -- this way we are not losing any data and are not making more copies with the same name.
* Make the copies have numbers but not randomized.   Maybe they can be in a progression.  


I really recommend leaving the random number thing in and then we can create a separate PR that tackles the core of the problem which is "How do we support creating copies without having orphan copies that bloat our system?"  
That seems like a very good chunk of problem to solve in isolation rather than squeezing it into this one.

Opoku-Agyemang (changes requested):

BradHN1:
Went with the first option.

src/api/store/vendor.py
BradHN1: I've implemented the first method.

src/api/store/event.py
BradHN1: All tests work now

Issue #448, opened 2022-03-03
Saved numbers for reported actions get set to crazy values.


BradHN1:
Seen a couple of times recently, it happened to Wayland today.
![image (5)](https://user-images.githubusercontent.com/12800484/156650961-0e506322-a52b-4ec5-bbdb-8a734ed4c416.png)

BradHN1:
Interim fix - add checks to updating of Data model values in the three locations it is added:
Check that value for actions in a category is less than the action total target value.
If not, send slack message to super_admin-notify channel

PR #452, opened 2022-03-09
Data corruption error debugging + 2 minor fixes

BradHN1:
Sam - I tried to make a Slack webhook to send messages to a Super Admin Notify channel.  Got a message from Slack that the public webhook created was disabled for security reasons.  I want to get a notification with data gets corrupted which is happening pretty often, so I can know which part of code is doing it.


src/api/services/utils.py
```diff
@@ -4,6 +4,11 @@
 
 
 def send_slack_message(webhook, body):
-  r = requests.post(url = webhook, data = json.dumps(body)) 
+  # fix for sending to Super Admin webhook.
+  # it needs to have a text field, so just stick the json string in there
+  if not body.get("text",None):
+    body["text"] = json.dumps(body)
+
+  r = requests.post(url = webhook, data = json.dumps(body))
```
Opoku-Agyemang, line 12: How did you create the webhook?  Is it a webhook created from Workflow Builder?  If not check this out https://slack.com/help/articles/360041352714-Create-more-advanced-workflows-using-webhooks

src/api/store/graph.py
```diff
@@ -393,3 +405,44 @@ def list_graphs_for_super_admin(self, context: Context):
     except Exception as e:
       capture_message(str(e), level="error")
       return None, CustomMassenergizeError(str(e))
+
+  def debug_data_fix(self) -> None:
+    # attempting to fix the problem with data getting screwed up
```
Opoku-Agyemang, line 410: I would love a documentation on what the problem is and what approach we are taking to solve it.  Its not very clear why we are doing what we are doing here.

Opoku-Agyemang (changes requested):

src/api/store/graph.py
```diff
@@ -393,3 +405,44 @@ def list_graphs_for_super_admin(self, context: Context):
...
+          tag = d.tag
+
+          for user_action in user_actions:
+            if user_action.action and user_action.action.tags.filter(pk=tag.id).exists():
+              val += 1
+
+          if (val != d.value) :
+            d.value = val
+            d.save()
+            print("WARNING - data_fix: Community: " + community.name
```
Opoku-Agyemang, line 444: maybe we don't want to commit print statements as they will be dumped in AWS logs which is virtually inaccessible to us.

src/api/services/utils.py
BradHN1: I set it up according to instructions onlin through an App i created called SuperAdminNotify.
The webhook you set up for a community admin channel doesn't work.

src/api/store/graph.py
BradHN1: API issue #448 documents this (https://github.com/massenergize/api/issues/448)
The reported values in the Data model get set to many thousands, so the impact page looks crazy.
We have observed this several times.  The debug_data_fix code is what I use to automatically fix it by running the portal locally, it takes a few seconds.  Works sort of like the data.backfill but much faster.

PR #453, opened 2022-03-10
Issue 449 - email to users who RSVP for events

BradHN1:


BradHN1:
Here is an example email sent from the back end:

[Thank you for registering for Your Sustainable Future Home Now.pdf](https://github.com/massenergize/api/files/8226885/Thank.you.for.registering.for.Your.Sustainable.Future.Home.Now.pdf)

Opoku-Agyemang (approved):

src/api/store/event.py
```diff
@@ -257,6 +253,8 @@ def create_event(self, context: Context, args) -> Tuple[dict, MassEnergizeAPIErr
         if not community:
           return None, CustomMassenergizeError("Please provide a valid community_id")
 
+      print("before event .create()")
```
Davidjwkinghorn, line 256: Not super important, but these print debugs should be removed when this PR gets merged.

src/database/models.py
```diff
@@ -1609,18 +1612,18 @@ def info(self):
 
     def simple_json(self):
         data = model_to_dict(
-            self, exclude=["tags", "image", "community", "invited_communities"]
+            self, exclude=["tags", "image", "community", "invited_communities", "user"]
         )
-        data["start_date_and_time"] = self.start_date_and_time
-        data["end_date_and_time"] = self.end_date_and_time
+        #data["start_date_and_time"] = self.start_date_and_time
```
Davidjwkinghorn, line 1617: Start date/time and end date/time for events should be included in simple_json, I think.

Davidjwkinghorn (changes requested):
Looks good to me apart from the two very minor changes I mentioned.

src/database/models.py
BradHN1: I agree, but they are already included by virtue of the model_to_dict call right above it.

Davidjwkinghorn: Okay, I was wondering about that. Sounds good then.

src/api/store/event.py
BradHN1: right you are.  I've removed them

src/database/models.py
BradHN1: Removed those commented lines

Davidjwkinghorn:
There's still a failed test, but it's something with teams, so it's outside the scope of this PR.

Davidjwkinghorn (approved):
LGTM

PR #465, opened 2022-03-23
Fix tests for github action

Opoku-Agyemang:
The goal of this PR is to make the github action that runs tests be configured well enough to run the tests we have.
If we dont fix this, with our rapid development, we could commit some bugs without realizing.

Also some of our tests were creating files and storing them in our s3 bucket each time we run our tests 😂  so i fixed that. Whew 😌 !

Strategy:
I introduce a new run mode called `test`.  When the api is running in test mode, it will create and use a sqlite file as the db instead of trying to connect to a remote host.
Its faster this way and allows the github action to run tests for us without having to provide a remote db access.

### DIFF
--- a/.github/workflows/tests-dev.yml
+++ b/.github/workflows/tests-dev.yml
@@ -21,7 +21,7 @@ jobs:
         pip install -r src/requirements.txt
     - name: Run Tests
       run: |
-        cd src && make test
+        make test --directory src
       env:
         SECRET_KEY: ${{ secrets.DEV_SECRET_KEY }}
         AWS_ACCESS_KEY_ID: ${{ secrets.DEV_AWS_ACCESS_KEY_ID }}
--- a/.github/workflows/tests-dev.yml
+++ b/.github/workflows/tests-dev.yml
@@ -21,7 +21,7 @@ jobs:
         pip install -r src/requirements.txt
     - name: Run Tests
       run: |
-        python src/manage.py test
+        cd src && make test
       env:
         SECRET_KEY: ${{ secrets.DEV_SECRET_KEY }}
         AWS_ACCESS_KEY_ID: ${{ secrets.DEV_AWS_ACCESS_KEY_ID }}
@@ -35,6 +35,7 @@ jobs:
         DATABASE_PASSWORD: ${{ secrets.DEV_DATABASE_PASSWORD }}
         DATABASE_HOST: ${{ secrets.DEV_DATABASE_HOST }}
         DATABASE_PORT: 5432
+        DJANGO_ENV: ${{ secrets.DJANGO_ENV }}
         EMAIL: ${{ secrets.DEV_EMAIL }}
         EMAIL_PASSWORD: ${{ secrets.DEV_EMAIL_PASSWORD }}
         EMAIL_PORT: 527
--- a/src/Makefile
+++ b/src/Makefile
@@ -78,6 +78,12 @@ export TEST_INPUTS=$(infile)
 export TEST_OUTPUTS=$(outfile)
 export GENERATE_INPUTS=$(generate_inputs)
 
-test: migrations
+test:
+	if [[ -f "test.sqlite3" ]]; \
+	then \
+		rm test.sqlite3; \
+	fi
+	touch test.sqlite3
+	DJANGO_ENV=test python3 manage.py migrate
 	DJANGO_ENV=test python3 -W ignore manage.py test $(package)
 .PHONY: test
--- a/src/Makefile
+++ b/src/Makefile
@@ -9,13 +9,13 @@ init:
 env=remote
 export DJANGO_ENV=$(env)
 start:
-	python deployment/prepare_to_deploy.py local 1
-	python manage.py runserver
+	python3 deployment/prepare_to_deploy.py local 1
+	python3 manage.py runserver
 .PHONY: start
 
 
 prepare-prod:
-	python deployment/prepare_to_deploy.py --prod 0 1
+	python3 deployment/prepare_to_deploy.py --prod 0 1
 .PHONY: prepare-prod
 
 VERSION=latest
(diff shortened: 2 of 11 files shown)


.github/workflows/tests-dev.yml
```diff
@@ -35,6 +35,7 @@ jobs:
         DATABASE_PASSWORD: ${{ secrets.DEV_DATABASE_PASSWORD }}
         DATABASE_HOST: ${{ secrets.DEV_DATABASE_HOST }}
         DATABASE_PORT: 5432
+        DJANGO_ENV: ${{ secrets.DJANGO_ENV }}
```
Opoku-Agyemang, line 38: setting this here.  I put the actual value in the github secrets.

src/Makefile
```diff
@@ -9,13 +9,13 @@ init:
 env=remote
 export DJANGO_ENV=$(env)
 start:
-	python deployment/prepare_to_deploy.py local 1
-	python manage.py runserver
+	python3 deployment/prepare_to_deploy.py local 1
```
Opoku-Agyemang, line 12: changing all these to python3 for consistency of all our developers.  Also it makes it easier for the github action to find the right python version we want and hence the right pip packages.

src/Makefile
```diff
@@ -57,11 +57,16 @@ deploy-dev:
...
+	python3 deployment/prepare_to_deploy.py --dev
 	docker-compose build
 	docker-compose push
 .PHONY: deploy
 
+.PHONY: migrations
+migrations: 
+	python3 manage.py makemigrations
+	python3 manage.py migrate
+
```
Opoku-Agyemang, line 69: adding this so that i can just do `make migrations` and have everything done for me

src/Makefile
```diff
@@ -73,6 +78,12 @@ export TEST_INPUTS=$(infile)
...
-test:	
-	python -W ignore manage.py test $(package)
+test:
+	if [[ -f "test.sqlite3" ]]; \
+	then \
+		rm test.sqlite3; \
+	fi
+	touch test.sqlite3
+	DJANGO_ENV=test python3 manage.py migrate
+	DJANGO_ENV=test python3 -W ignore manage.py test $(package)
```
Opoku-Agyemang, line 88: adding a test command.
First it checks if the test.sqlite3 file is present then deletes it to create a fresh start before running the tests.

src/_main_/settings.py
```diff
@@ -178,8 +180,14 @@
         'HOST'     : os.environ.get('DATABASE_HOST'),
         'PORT'     : os.environ.get('DATABASE_PORT')
     },
+    'test_db': {
+        "ENGINE": "django.db.backends.sqlite3",
+        "NAME": 'test.sqlite3',
+    }
```
Opoku-Agyemang, line 186: our test db should always be the sqlite3 file

src/_main_/settings.py
```diff
@@ -178,8 +180,14 @@
...
 }
 
+if is_test_mode():
+    DATABASES['default'] = DATABASES['test_db']
```
Opoku-Agyemang, line 190: if we are in test mode, then use the test db.

src/_main_/utils/emailer/send_email.py
```diff
@@ -5,10 +5,14 @@
 import pystmark
 
 from _main_.settings import EMAIL_POSTMARK_SERVER_TOKEN
+from _main_.utils.utils import is_test_mode
 
 FROM_EMAIL = 'no-reply@massenergize.org'
 
 def old_send_massenergize_email(subject, msg, to):
+  if is_test_mode():
+    return True
```
Opoku-Agyemang, line 14: when we are in test mode, we should not actually be sending emails.

src/api/store/graph.py
```diff
@@ -407,39 +407,42 @@ def list_graphs_for_super_admin(self, context: Context):
...
-        action_goal= max(community.goal.target_number_of_actions, 100)
-      else:
-        # communities that don't have goals
-        action_goal = 100
-
-      if community.is_geographically_focused:
-        user_actions = UserActionRel.objects.filter(
-          real_estate_unit__community=community, status="DONE"
-        )
-      else:
```
Opoku-Agyemang, line 422: this function was breaking whenever migrations hadn't been done.  I surrounded it with a try catch.

src/carbon_calculator/CCDefaults.py
```diff
@@ -120,6 +122,7 @@ def exportDefaults(self,fileName):
             csvfile.close()
         return status
     def importDefaults(self,fileName):
+        csvfile = None
```
Opoku-Agyemang, line 125: when something goes wrong in the function, the finally clause uses the csvfile variable which is not declared.  I am initializing it here so that the code does not break

src/_main_/settings.py
```diff
@@ -114,8 +116,9 @@
 
 
 #-------- FILE STORAGE CONFIGURATION ---------------------#
-DEFAULT_FILE_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
-STATICFILES_STORAGE  = 'storages.backends.s3boto3.S3Boto3Storage'
+if not is_test_mode():
+    DEFAULT_FILE_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
+    STATICFILES_STORAGE  = 'storages.backends.s3boto3.S3Boto3Storage'
```
Opoku-Agyemang, line 121: we should never be using s3 if we are running tests!

src/Makefile
Davidjwkinghorn: 100%!

src/_main_/utils/emailer/send_email.py
Davidjwkinghorn: I agree. If we need to test emails we can send to the Postmark test server instead. That should be as simple as swapping the API keys.

sopokuagyemang (approved):

Davidjwkinghorn (approved):
Looks good to me!

PR #474, opened 2022-04-06
473 - back end changes for guest user sign-in

BradHN1:
New route auth.sigininasguest
SImple changes to users.create

### DIFF
--- a/src/api/handlers/auth.py
+++ b/src/api/handlers/auth.py
@@ -4,12 +4,15 @@ from _main_.utils.route_handler import RouteHandler
 from _main_.utils.common import parse_list, parse_bool, check_length, rename_field
 from api.services.auth import AuthService
 from _main_.utils.massenergize_response import MassenergizeResponse
+from _main_.utils.massenergize_errors import NotAuthorizedError
 from types import FunctionType as function
 from _main_.utils.context import Context
 from _main_.utils.validator import Validator
 from api.decorators import admins_only, super_admins_only, login_required
 from _main_.settings import RUN_SERVER_LOCALLY
 
+ONE_YEAR = 365*24*60*60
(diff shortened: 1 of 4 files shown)


frimpongopoku (approved):
* As I have expressed in many places, I am not a fun of entagling things like this. Our frontend has taught me to prevent that as much as possible 😅  . 
But hey, it seems like this would work in the short term, as long as we will structure things properly when earth day is over. 
 so "LGTM"? 🤣 

Also, @BradHN1  have we catered for the scenario where  I hit the `users.create` route with a guest email and the email is already used by: 

* A proper user in the system 
* Or The same guest? 

Will it send an `error` response? We dont want that right? My understanding is that we want the process to be as smooth as possible right? So in that case should we return the already existing user?  (At least for an already existing guest)

BradHN1:
Good point.  I appreciate your concern, but I think this will work fine in the long term.

The question as to an existing user trying to sign in as guest is good.  We don’t want people with profiles to sign in as guest, but to use their profile.  Ideally this won’t happen too often since once signed in you are signed in for a year, but when changing devices or browsers it will certainly happen.  Right now it will return an error not_a_guest_user, and if you see this you can ask the user to sign in with their profile.  We’ll have to talk this through.

frimpongopoku:
Alright. Just a note, the long term and short term I am talking about here is about the implementation -- how it's being implemented in the code. As we move and and we are met with different scenarios which call for different ways of giving users access, things will get jumbled up in there, and we would need better ways of doing things(but I guess we will just fix and restructure as they come 🤣 🤣 ). Also the entire logic of differentiating a proper registration from a guest registration should definitely not hinge on the user's full name having "guest" you see.... Yeah, that is what I was thinking.

PR #475, opened 2022-04-10
Updates to users.create

BradHN1:
Don't automatically turn a standard user instead of a guest user

### DIFF
--- a/src/api/store/userprofile.py
+++ b/src/api/store/userprofile.py
@@ -14,7 +14,7 @@ import json
 from typing import Tuple
 from api.services.utils import send_slack_message
 from _main_.settings import SLACK_SUPER_ADMINS_WEBHOOK_URL
-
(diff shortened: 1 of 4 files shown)


src/api/store/userprofile.py
```diff
@@ -371,9 +371,16 @@ def create_user(self, context: Context, args) -> Tuple[dict, MassEnergizeAPIErro
       # added for special case of guest users, mark them as such in user_info
       user_info = args.get('user_info', None)     
       full_name = args.get('full_name')
+      preferred_name = args.get('preferred_name', None)
+
+      new_user_type = STANDARD_USER
       if 'guest' in full_name.lower():
```
frimpongopoku, line 377: @BradHN1  let me add `is_guest:true` to the api request when I am hitting the `users.create` route. Instead of using the name to differentiate standard user from guest user yeah? 
You can expect `is_guest` in the handler, but dont make it compulsory.

frimpongopoku (approved):

PR #476, opened 2022-04-12
responsive landing page

frimpongopoku:
Found that our landing page looked a bit off on mobile. So I just cleared it up a bit 

Was like this: 
Screenshot 2022-04-12 at 12 55 22

Now its like this: 

Screenshot 2022-04-12 at 12 54 10

### DIFF
--- a/src/website/templates/communities.html
+++ b/src/website/templates/communities.html
@@ -1,142 +1,136 @@
-{% extends "base.html" %}
-{% block javascript %}
+{% extends "base.html" %} {% block javascript %}
 <script>
-        function onCommunitySelected(selectElement){
-            const subdomain = selectElement.value;
-            const suffix = selectElement.sandbox ? "?sandbox=true" : "";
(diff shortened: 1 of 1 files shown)


BradHN1:
I like it.  However, like many people, my eyes are not as good as they were once.  So I would ask that you increase the text size in two places:
1. IN the Welcome to our Community Portal, make the font size closer to what it was, but aligned centered, with the line broken between "our" and "Community"
2. On the FInd Out How To, keep font size as it was, but center aligned.

frimpongopoku:
> I like it. However, like many people, my eyes are not as good as they were once. So I would ask that you increase the text size in two places:
> 
> 1. IN the Welcome to our Community Portal, make the font size closer to what it was, but aligned centered, with the line broken between "our" and "Community"
> 2. On the FInd Out How To, keep font size as it was, but center aligned.

Is this better? @BradHN1 
Screenshot 2022-04-12 at 16 51 12

PR #482, opened 2022-04-20
Only send Welcome email once,

BradHN1:
Send welcome email when a user profile updated to accept terms and conditions.  Previously this could happen multiple times.


Davidjwkinghorn (approved):
LGTM

PR #508, opened 2022-05-18
Cadmin replies saved to DB

abdullai-t:
####  Summary / Highlights
This PR focuses on saving cadmins reply to the database.

#### Details (Give details about what this PR accomplishes, include any screenshots etc)
The above was achieved by recycling the `Message model` and adding a field named `parent` which is set to the message instance been replied to. A normal message will have its `parent field ` to be null and a reply would a non-null `parent field`.

* [x] I've given my PR a meaningful title

### DIFF
--- a/src/api/services/message.py
+++ b/src/api/services/message.py
@@ -42,7 +42,10 @@ class MessageService:
       title = args.pop('title', None)
       to = args.pop('to', None)
       body = args.pop('body', None)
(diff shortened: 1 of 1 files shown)


BradHN1 (approved):
LGTM
I added a couple of additions to fix the format of the reply message, and return the serialized new message rather than a boolean.

PR #509, opened 2022-05-19
Matching changes to fit frontend Media Library Applications

frimpongopoku:
Made changes to 
- [X] Creating  & Editing Actions 
- [X] Creating & Editing Events 
- [X] Creating & Editing Teams 
- [X] Creating & Editing Vendors 
- [X] Creating & Editing Testimonials 
- [X] Creating new communities 
- [X] Editing community homepage information

Before, these routes used to expect a media file, which was uploaded and used to create a new media object and then attached to (event, actions... etc whatever the scenario was). 
Now with the media library, I just  pass an array of ids from the frontend instead of media file(s). Ids of media objects that exist already. So the new logic is just to find the media objects that the admins have chosen, and just attach them to the use case....


src/api/store/team.py
```diff
@@ -292,17 +293,13 @@ def update_team(self, context, args) -> Tuple[dict, MassEnergizeAPIError]:
...
-          new_logo = len(logo) > 1000
-          if new_logo:
-            logo = Media.objects.create(file=logo, name=f"{slugify(team.name)}-TeamLogo")
-            logo.save()
-            team.logo = logo
-          else: 
-            team.logo = None 
+     
+      if logo: #now, images will always come as an array of ids, or "reset" string 
+        if logo[0] == "reset": #if image is reset, delete the existing image
```
abdullai-t, line 298: Would be great if you could turn the `'reset'` thing into some variable

src/api/store/action.py
```diff
@@ -189,10 +190,12 @@ def update_action(self, context: Context, args) -> Tuple[dict, MassEnergizeAPIEr
       action.update(**args)
       action = action.first()
 
-      # If no image passed, don't delete the existing
-      if image:
-        media = Media.objects.create(name=f"{action.title}-Action-Image", file=image)
-        action.image = media
+      if image: #now, images will always come as an array of ids, or "reset" string 
+        if image[0] == "reset": #if image is reset, delete the existing image
```
abdullai-t, line 194: here too

src/api/store/team.py
Opoku-Agyemang: + 1

Define a constant up there and reference it here

Opoku-Agyemang (approved):

src/api/handlers/action.py
```diff
@@ -133,7 +133,7 @@ def update(self, request):
       .expect("action_id", int, is_required=True)
       .expect("title", str, is_required=False, options={"min_length": 4, "max_length": 40})
       .expect("calculator_action", int, is_required=False)
-      .expect("image", "file", is_required=False, options={"is_logo": True})
+      .expect("image", "str_list", is_required=False, options={"is_logo": True})
```
Opoku-Agyemang, line 136: with this are you sure things will not go kaboom? 😂 
Are we first uploading the images somewhere and getting links?  Can the links expire?

src/api/store/action.py
```diff
@@ -77,8 +78,8 @@ def create_action(self, context: Context, args) -> Tuple[dict, MassEnergizeAPIEr
         community = Community.objects.get(id=community_id)
         new_action.community = community
       
-      if image:
-        media = Media.objects.create(name=f"{args['title']}-Action-Image", file=image)
+      if image: #now, images will always come as an array of ids 
+        media = Media.objects.filter(pk = image[0]).first()
```
Opoku-Agyemang, line 82: how are we ensuring backward compatibility?

src/api/store/vendor.py
```diff
@@ -76,8 +76,7 @@ def create_vendor(self, context: Context, args) -> Tuple[Vendor, MassEnergizeAPI
 
       new_vendor = Vendor.objects.create(**args)
       if image:
-        logo = Media(name=f"Logo-{slugify(new_vendor.name)}", file=image)
-        logo.save()
+        logo = Media.objects.filter(pk = image[0]).first()
```
Opoku-Agyemang, line 79: are we asserting that at least one image is present? and that it is a list?  Also if its an array of id's maybe the word "image" is a tad bit misleading

Opoku-Agyemang (changes requested):
good job but a few nit comments

src/api/store/vendor.py
frimpongopoku: Yes, the array will never be empty, and will always have one item at least. In scenarios where we dont want to do anything with images, we do not provide the "image" field at all in the request payload. So this never triggers.

src/api/store/action.py
frimpongopoku: We are not! I never touched the old uploader on the frontend side. I just added the media library as a new field. So, devs can still create forms with the older uploader, or the media library. And when they do, they will need to handle things on the backend in a way that matches the outputs of the fields... I just thought there is  never going to be a scenario where we are expecting one field to come as the kind of data we are expecting from the media library(array of ids) or the old uploader(file Object) -- its always going to be one way, and it will be one we are prepared for isnt it?

frimpongopoku:
I think I am done here as well.

Issue #412, opened 2022-01-05
Platform Users should have unique user name to enable communication


BradHN1:
Currently, users are asked for First Name, Last Name, and Preferred Name.

Often they enter a very short preferred name, such that it wouldn't be possible to tell users apart.
In that case, it will cause problems when it becomes possible for users to communicate with others.

This could be fixed by requiring the Preferred Name to be unique, going through and adding a suffix to each, and inviting users to verify that is ok. 

Instead we could take the model of Twitter, create a new field (a "handle"), which we create automatically from @ + First Initial + Last Name + number, and invite user back to platform to change it if they like.  Handles could be changed, but would need to be  non-reusable (except by the same user). 

Once this is implemented in the API, would need to implement display of handle together with name and profile pic on team page and wherever else needed.

Davidjwkinghorn:
We will be replacing "Preferred name" with "User name". When we convert to unique usernames we'll have to do the following as well:
- [ ] Notify users that usernames are now unique, but that they can change their username later. (This is fine as long as users are traceable back to userIDs. In cases of abuse/harassment etc, it's important to be able to know who's doing it. If they're able to escape detection by changing their username that's a very real issue.)
- [ ] Automatically change existing "Profile names" where duplicates exist and make the "User names" a unique version of what they had before.

PR #528, opened 2022-07-01
just one API call

mattia-danese:
####  Summary / Highlights

Made so PR #412 only needs to make one API call per username suggestion/validation.


src/api/store/userprofile.py
```diff
@@ -1,3 +1,4 @@
+from tkinter import N
```
BradHN1, line 1: Remove this...

BradHN1 (changes requested):
Looks good - I will test locally
Lets add unit tests to test_userprofile.py for the route you added.

src/api/store/userprofile.py
```diff
@@ -132,11 +133,38 @@ def __init__(self):
...
+        if not UserProfile.objects.filter(preferred_name=username).exists():
+            return [True, username], None
+
+        # username exists, finds next available closest username
+        usernames = list(UserProfile.objects.filter(preferred_name__istartswith=username).order_by('preferred_name').values_list("preferred_name", flat=True))
+
+        if len(usernames) == 1:
+            suggestion = username + "-0"
+            return [False, suggestion], None
+        numbers = [int(x[x.index("-")+1:]) for x in usernames[1:]]
```
BradHN1, line 151: Would this expression be a problem if someone had a "-" in the middle of their name?

BradHN1: Testing on Dev with the frontend-portal changes, the usernames returned when I tried to use BradHN were:
['BradHN', 'BradHN-0', 'BradHN24', 'BradHN5', 'BradHN-on-dev', 'BradHN-testing-!']

src/api/tests/test_userprofile.py
```diff
@@ -476,11 +476,13 @@ def test_validate_username(self):
...
         self.assertTrue(response["success"])
-        self.assertEqual(response["username"], "thisUsernameDoesNotExistAlready")
+        self.assertTrue(response["data"]["valid"])
+        self.assertEqual(response["data"]["suggested_username"], "thisUsernameDoesNotExistAlready")
 
         # invalid (not unique) username
-        response = self.client.post('/api/users.validate.username', urlencode({'username': "Kaat from AnyTown"}), content_type="application/x-www-form-urlencoded").toDict()
-        self.assertFalse(response["success"])
-        self.assertEqual(response["username"], "Kaat from AnyTown-0")
+        response = self.client.post('/api/users.validate.username', urlencode({'username': "BradHN"}), content_type="application/x-www-form-urlencoded").toDict()
```
BradHN1, line 483: The test should use self.USER.preferred_name since that is a user created at the start of testing.  BradHN doesn't exist in the test database.

src/api/tests/test_userprofile.py
```diff
@@ -476,11 +476,13 @@ def test_validate_username(self):
...
+        self.assertTrue(response["success"])
+        self.assertFalse(response["data"]["valid"])
+        self.assertEqual(response["data"]["suggested_username"], "BradHN1")
```
BradHN1, line 486: ditto - check that suggested_username is self.USER.perferred_name+"1"

BradHN1 (approved):

PR #533, opened 2022-07-06
Change user.delete route to delete user completely

BradHN1:
Splitting an earlier PR into two: this part is the changes to delete a user completely, and changes so that the deleted user doesn't show up in any queries.


src/api/store/userprofile.py
```diff
@@ -563,18 +563,77 @@ def update_user(self, context: Context, args) -> Tuple[dict, MassEnergizeAPIErro
...
-        if context.user_id != user_id:
-          return None, NotAuthorizedError()
+      if not self._has_access(context, user_id):
+        return None, CustomMassenergizeError("permission_denied")
       
       users = UserProfile.objects.filter(id=user_id)
-      users.update(is_deleted=True)
+      user = users.first()
+      old_email = user.email
+      new_email = "DELETED-" + datetime.today().strftime('%Y%m%d-%H%M') + "-" + old_email
```
Opoku-Agyemang, line 572: why are we changing the email?

src/api/store/userprofile.py
```diff
@@ -563,18 +563,77 @@ def update_user(self, context: Context, args) -> Tuple[dict, MassEnergizeAPIErro
...
+
+      # if a Team includes on Admins, remove it and notify other admins. if no other admins notify cadmin
+      teams = user.team_admins.filter(is_deleted=False)
+      for team in teams:
+        team.admins.remove(user)
+        # Don't bother with the notify
+        # if team.admins.count() == 0:
+        #  community_admin = team.primary_community.owner_email
+        #  msg = "User %s has been deleted, and Team %s no longer has a Team Admin" % (old_email, team.name)
+        #  cadmin_messages.append(msg)
```
Opoku-Agyemang, line 602: remove if we don't need it

src/carbon_calculator/calcUsers.py
```diff
@@ -30,7 +30,8 @@ def QueryAllCalcUsers(args):
     event = args.get('Event','')
     group = args.get('Group','')
     if event == '':
-        users = UserProfile.objects.all()
+        #users = UserProfile.objects.all()
```
Opoku-Agyemang, line 33: remove

src/api/store/userprofile.py
BradHN1: Since we do not delete the record from the database, we modify the email address in case the person wants to create a profile again with that email.  This way we can tell exactly what happens in case needed.

src/api/store/userprofile.py
BradHN1: done

src/api/store/userprofile.py
sopokuagyemang: we should leave a comment on that line describing exactly what you just said here

BradHN1: Done. Thanks for the request

Opoku-Agyemang (approved):

PR #534, opened 2022-07-06
Add is_approved flag to Action/Event/Vendor models

BradHN1:

* [x] I moved the linked issue to "QA Verification" now that it is ready to merge.


src/database/models.py
```diff
@@ -1110,7 +1110,6 @@ def full_json(self):
         data = self.simple_json()
         # Q: should this be in simple_json?
         data["communities"] = [c.simple_json() for c in self.communities.all()]
-        # data["admins"] = [a.simple_json() for a in self.admins.all()]
```
Opoku-Agyemang, line 1113: any reason we are removing this?

Opoku-Agyemang (commented):
How will this also affect the queries we make to actions.list etc?  Will is_approved be used to filter?

src/database/models.py
BradHN1: Because it is a comment and the admins are returned by the simple_json method.

BradHN1:
@Opoku-Agyemang 

> How will this also affect the queries we make to actions.list etc? Will is_approved be used to filter?
Good question.  Right now those queries work the same way as they do for Testimonials, which is that is_approved is ignored and only items with is_published=True are returned unless you are in sandbox. 

 Also right now the 'is_approved' is not required to be set true for is_published to be set to true; I am going to fix this so can hold off on this PR for the moment.

Davidjwkinghorn (approved):
LGTM

PR #541, opened 2022-07-21
tags in actions will get copied over, and tweaks to action serializer

frimpongopoku:
Related Ticket :  https://github.com/massenergize/api/issues/542

- [x] Tags on actions will get copied over when copying an action 
- [x] Also found that creating action with media library was not selecting the right image, fixed that
- [x] Few mods to actions model to show vendors in simple_json

https://user-images.githubusercontent.com/26961591/180138783-510d3392-0e66-4423-a21f-f6bf7127e371.mov


BradHN1 (approved):
LGTM - tested locally

PR #551, opened 2022-07-29
media lib api mods to allow more info

frimpongopoku:
Tiny model tweaks to allow media library to keep, and show more info about image. Like ( size of image, and who uploaded). 

Models were tweaked a bit, so you would need to makemigrations and migrate

Related to frontend API: https://github.com/massenergize/frontend-admin/pull/668


BradHN1:
@frimpongopoku - one more feature request:
in the information side pane for an image, it shows which content it is used in, in blue text.  Can those be links, so that when I click on one it opens a new tab with that content editor?

frimpongopoku:
The content of this PR have been added to this new PR instead: https://github.com/massenergize/api/pull/585

Issue #553, opened 2022-04-21
Issue adding CAdmin on community creation.


Davidjwkinghorn:
Ellen:
> found bug this am, when kaat set up a new community. It auto added the person we listed on the footer as an admin, but when he went to sign in as a Cadmin it did not work. Kaat then removed him and readded him as a admin -- Kaat can weigh in on the fix, fine if person on footer is auto made an admin and we can always delete them but it does need to allow them to function as the admin

PR #554, opened 2022-08-03
tiny tweak to allow primary contact to be admin automatically

frimpongopoku:
Related to ticket #553  

This change will now make primary contact an admin of the community, as described in the ticket


BradHN1 (changes requested):
I think this only does part of the job.
To record the owner as a community admin of the community for which he/she is the owner, you need to add them to the CommunityAdminGroup.  See add_community_admin in admin.py for details.

frimpongopoku:
> I think this only does part of the job. To record the owner as a community admin of the community for which he/she is the owner, you need to add them to the CommunityAdminGroup. See add_community_admin in admin.py for details.

Hi @BradHN1  True, but that part of the code is already implemented. The part I am dealing with is when a super admin first creates a community, and wants to add the primary contact as an automatic admin. Whoever implemented this was actually on track, so they already set the primary contact up in the community admin group, they just forgot that the `_is_community_admin` field of the user object also has to be turned on. Check `store/community.py` `create_community` `line 704` you will see that the user is already added to  a communityAdminGroup.

frimpongopoku:
Jeez what happened to this PR? 🤣

PR #566, opened 2022-08-18
tiny tweak to allow primary contact to be admin automatically

### LINKED ISSUE #553: Issue adding CAdmin on community creation.

frimpongopoku:
Related to ticket #553  

This change will now make primary contact an admin of the community, as described in the ticket. 

- User is already added  into communityAdminGroup already, so we just had to flip the is_admin property

closes #553

### DIFF
--- a/src/api/store/community.py
+++ b/src/api/store/community.py
@@ -714,7 +714,9 @@ class CommunityStore:
 
             owner_email = args.get("owner_email", None)
             if owner_email:
-                owner = UserProfile.objects.filter(email=owner_email).first()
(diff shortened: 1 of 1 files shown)


BradHN1 (approved):
LGTM

PR #568, opened 2022-08-19
567 - Fix FeatureFlag logic

BradHN1:
Make the feature flags work for EVERYONE, ALL_EXCEPT and SPECIFIED communities and users.

### DIFF
--- a/src/database/models.py
+++ b/src/database/models.py
@@ -442,9 +442,16 @@ class Community(models.Model):
                 l = loc.zipcode
             locations += l
 
-
-        features_flags = FeatureFlag.objects.filter(communities=self)
-        features_flags_json = [f.simple_json() for f in features_flags]
+        # Feature flags can either enable features for specific communities, or disable them
+        feature_flags = FeatureFlag.objects.all()
+        feature_flags_json = []
+        for f in feature_flags:
+          specified_communities = f.communities.all()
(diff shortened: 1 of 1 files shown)


abdullai-t (approved):
LGTM ✅

src/database/models.py
```diff
@@ -442,9 +442,16 @@ def full_json(self):
                 l = loc.zipcode
             locations += l
 
-
-        features_flags = FeatureFlag.objects.filter(communities=self)
-        features_flags_json = [f.simple_json() for f in features_flags]
+        # Feature flags can either enable features for specific communities, or disable them
+        feature_flags = FeatureFlag.objects.all()
```
Opoku-Agyemang, line 446: why are we fetching all feature flags here?  I really dont think this logic should live here.  We have a separate endpoint to fetch feature flags based on some specific filters.  Why not use that?

Opoku-Agyemang (changes requested):

BradHN1:
There will always be a very small number of feature flags.  And we need ti to work this week, instead of being completely broken.

src/database/models.py
BradHN1: We need this working now, and it was broken before., and the time penalty will be very small for loading a community.  There are many much slower things to worry about.

PR #585, opened 2022-09-16
Tags to media library

frimpongopoku:
Works with frontend PR: https://github.com/massenergize/frontend-admin/pull/718 

- [x] Changes to allow admins to add tags to images when they upload 
- [x] Changes to allow admins to include tags as a search criteria 
- [x] Pulled in all the changes in [this PR](https://github.com/massenergize/api/pull/551) as well.


src/api/services/media_library.py
```diff
@@ -68,13 +68,13 @@ def getImageInfo(self, args):
...
-        media_json = get_json_if_not_none(media)
+        media_json = get_json_if_not_none(media, True)
         return {
             **media_json,
-            "user_info": user_info,
-            "info": {
-                "events": events,
-                "actions": actions,
-                "testimonials": testimonials,
+            "information": user_info,
```
Opoku-Agyemang, line 74: i think user_info is more explicit.  information is very generic.  You can also go for `user_information`

Opoku-Agyemang (changes requested):
i think the tests are breaking.  Can you add tests for this new functionality and also make sure the other tests are not breaking?

Davidjwkinghorn:
I think I'm ready to merge this in, but I'm not seeing the filters for the tags anywhere when I run this with the admin portal changes. @frimpongopoku do you know why that might be?

Davidjwkinghorn:
I'm going to try merging this, in case it's just something weird with my local environment.

PR #595, opened 2022-09-28
updating events to make them live wont look for approval again

frimpongopoku:
Related to this ticket https://github.com/massenergize/frontend-admin/issues/747

So here is the DEMO of the bug @kaatvds  found
 

https://user-images.githubusercontent.com/26961591/192725202-badf7238-e76c-401a-a81d-b95bb158f7b0.mov

I dug into the api codebase, and found that we have now added a check during event update to check for approval...?? 

I wasnt sure what we needed approval for at first, since its admins who are in charge of creating events in the first place. But then I remembered, in one of our meetings, we discussed allowing user submitted content for (actions, events etc) so I guess thats what the part of the code was building towards.  

That said, I checked the frontend portal we have now, and did not see any interface that allows users to submit events yet 🤔 ...
I went back to the admin portal and also did not see any implementation that would help admins approve user submitted events either. 

We have also not put any checks in place on the backend to identify which events were submitted by users, and which ones are created by admins, to go ahead with this way of checking for approval... 

So in the end, I just had to comment part of the code out. It looks like its way too far ahead.  

And yes! Commenting out that part of the code, fixes the bug 💀... So, yipeee! 👯 

(Let me know if I am missing anything 🤔 )


kaatvds:
This is still happening: 
Screen Shot 2022-09-30 at 3 13 04 PM

Davidjwkinghorn:
I've made some changes to the logic, factoring in approval. Now when an event is published(made live) it's automatically approved. This should still hold up when we add event approval. @frimpongopoku can you check this over to see if it looks good?

Davidjwkinghorn:
Also something seems to be causing tests to fail. Something to do with the carbon calculator 🤷

Opoku-Agyemang (approved):

PR #611, opened 2022-12-06
Continue brads preference changes

frimpongopoku:
## Works with Admin frontend PR - https://github.com/massenergize/frontend-admin/pull/782

- [x] Event list for sharing will now show only live and future events 
- [x] Communities for sharing will also be only live communities 

NB: Automated tests will fail, cos the old tests dont cater for the new changes yet! ( Will be done soon)


BradHN1 (approved):
LGTM

PR #620, opened 2022-12-21
Fix a couple problems with the events nudge

BradHN1:
Two issues:
619 - problems with events nudge
also put one route get_footage behind admins_only


frimpongopoku (approved):
LGTM

PR #623, opened 2022-12-29
Filter, Search and Pagination

abdullai-t:
####  Summary / Highlights
See [document](https://tabdullai.notion.site/ME-PAGINATION-SEARCH-AND-FILTER-469085b3ac984435980baa0ffa56d8f2) for overview

This PR also fixes all broken tests


src/_main_/settings.py
```diff
@@ -27,8 +27,8 @@
 # ********  LOAD CONFIG DATA ***********#
```
BradHN1, line 27: Please avoid committing setting.py unless you have actual changes to it, not just for testing

src/api/handlers/action.py
```diff
@@ -230,3 +231,12 @@ def super_admin_list(self, request):
     if err:
       return err
     return MassenergizeResponse(data=actions)
+
+
+  @super_admins_only
+  def search_and_filter_actions(self, request):
```
BradHN1, line 237: it doesn't look like search_and_filter is fully implemented.  should it be part of this PR?

abdullai-t: the search and filter was implemented through the list items route

src/_main_/utils/pagination/__init__.py
```diff
@@ -0,0 +1,25 @@
+from django.core.paginator import Paginator, EmptyPage
+
+from _main_.utils.common import serialize_all
+PAGINATION_LIMIT = 50
```
BradHN1, line 4: could we pass the page size based on the Admin portal settings page length (can display up to 100 in the lists)

BradHN1:
@abdullai-t - Question about the format of the API JSON responses:
The list routes response structure has changed
`{"data": {"meta": {"next": 1, "count": 6}, "items": [{"id": 4, "name": "Just a Test Site",  ...]}, "error": null, "success": true}`

But other routes have not changed, for example communities.info:
`{"data": {"id": 6, "name": "Cooler Concord" ... }, "error": null, "success": true}`

Is it necessary to change the structure on the list routes, which forces version updates on all repositories at once (including ones which you don't use)?  Would it be ok for the list route JSON response be changed to:
`{"data": {"id": 4, "name": "Just a Test Site",  ...}, {"meta": {"next": 1, "count": 6}, "error": null, "success": true}`

src/_main_/settings.py
abdullai-t: sure

src/_main_/utils/pagination/__init__.py
```diff
@@ -0,0 +1,24 @@
+from django.core.paginator import Paginator, EmptyPage
+
+from _main_.utils.common import serialize_all
+
+def paginate(queryset, page, limit):
```
BradHN1, line 5: should we have default limit=50 if not specified?

src/api/services/action.py
```diff
@@ -28,7 +28,7 @@ def list_actions(self, context: Context, args) -> Tuple[list, MassEnergizeAPIErr
     actions, err = self.store.list_actions(context, args)
     if err:
       return None, err
-    return serialize_all(actions), None
+    return actions, None
```
BradHN1, line 31: What you've done is fine, but I think it would be slightly cleaner to have the pagination in the service layer routines.
So instead of:
return serialize_all(actions) 
replace with
return paginate(actions,page,limit)

abdullai-t: I will clean it up

src/_main_/utils/pagination/__init__.py
BradHN1: if limit not passed in args, this would be passed as null.  Let's protect against that with default=50

src/_main_/utils/common.py
```diff
@@ -109,6 +109,10 @@ def rename_fields(args, pairs):
 def serialize_all(data, full=False):
   if not data:
     return []
+
+  if type(data[0]) == dict:
+    return data
```
Opoku-Agyemang, line 114:
```suggestion
  if isinstance(data, dict):
    return data
```

Opoku-Agyemang: Is there a reason why you are doing data[0] ? what if the data is an int?

src/_main_/utils/pagination/__init__.py
```diff
@@ -0,0 +1,26 @@
+from django.core.paginator import Paginator, EmptyPage
+from _main_.utils.common import serialize_all
+
+DEFAULT_LIMIT = 50
```
Opoku-Agyemang, line 4: can this be in constants.py somewhere?

src/_main_/utils/pagination/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+def paginate(queryset, page, limit):
+    try:
+        if not limit:
+            limit = DEFAULT_LIMIT
+        p = Paginator(queryset, limit)
```
Opoku-Agyemang, line 9:
```suggestion
        paginator = Paginator(queryset, limit or DEFAULT_LIMIT)
```

src/_main_/utils/pagination/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+        items = []
+        pag = p.page(page)
```
Opoku-Agyemang, line 11: nit: the spelling
```suggestion
        current_page = p.page(page)
```

src/_main_/utils/pagination/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+        meta = {
```
Opoku-Agyemang, line 12: in pagination terms, this is usually called a cursor.
Use that instead

```suggestion
        cursor = {
```

src/_main_/utils/pagination/__init__.py
Opoku-Agyemang: Name your variables well to avoid confusion between page and paginator.  When someone sees `p` they could guess either.  Make it less ambiguous

src/_main_/utils/pagination/__init__.py
Opoku-Agyemang: is it the current_page? previous? next/. name it well

src/_main_/utils/pagination/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+        }
+        items = serialize_all(list(pag))
+        to_return = {
+            'meta':meta,
+            "items": items
+        }
+        
+        return to_return
+    except EmptyPage:
+        return p.get_page(1).object_list
```
Opoku-Agyemang, line 24: what does this do?  Leave a comment

src/api/handlers/action.py
```diff
@@ -113,15 +113,22 @@ def list(self, request):
...
 
     action_info, err = self.service.list_actions(context, args)
+
     if err:
       return err
-    return MassenergizeResponse(data=action_info)
+
+    meta = action_info.get('meta')
+    data = action_info.get('items')
+    return MassenergizeResponse(data=data, meta=meta)
```
Opoku-Agyemang, line 130: why dont you pass the data in and do this split in the MassenergizeREesponse itself?  then we will reduce a lot of the file changes even.

src/api/handlers/action.py
```diff
@@ -220,7 +227,9 @@ def community_admin_list(self, request):
     actions, err = self.service.list_actions_for_community_admin(context, args)
     if err:
       return err
-    return MassenergizeResponse(data=actions)
+    meta = actions.get('meta')
+    data = actions.get('items')
+    return MassenergizeResponse(data=data, meta=meta)
```
Opoku-Agyemang, line 232: see comment above

src/api/handlers/event.py
```diff
@@ -253,6 +253,7 @@ def list(self, request):
     self.validator.expect("community_id", is_required=False)
     self.validator.expect("subdomain", is_required=False)
     self.validator.expect("user_id", is_required=False)
+    self.validator.expect("limit", is_required=False)
```
Opoku-Agyemang, line 256: maybe consider extracting the `limit` and `cursor` or `meta` in the context middle layer.  And set it there.  that means by default you make the change in one place and can get it anywhere there is a context.

src/api/services/action.py
```diff
@@ -116,11 +118,14 @@ def list_actions_for_community_admin(self, context: Context, args) -> Tuple[list
     actions, err = self.store.list_actions_for_community_admin(context, args)
     if err:
       return None, err
-    return serialize_all(actions), None
+    sorted = sort_items(actions, context.args.get("params"))
```
Opoku-Agyemang, line 121: have a `get_params` method in context so you can just do `context.get_params()`

src/api/utils/filter_functions.py
```diff
@@ -0,0 +1,366 @@
+import json
+import operator
+from functools import reduce
+from django.db.models import Q
+
+
+def get_events_filter_params(params):
```
Opoku-Agyemang, line 7: these queries look so complicated and expensive.  We will have to review them thoroughly.

Opoku-Agyemang (changes requested):
I really would have loved for you to complete for one thing.  Like for Actions or something.  So that we can review that thoroughly then you can replicate for the others.  

Anyways i left a few comments

src/api/handlers/action.py
```diff
@@ -220,7 +224,7 @@ def community_admin_list(self, request):
     actions, err = self.service.list_actions_for_community_admin(context, args)
     if err:
       return err
-    return MassenergizeResponse(data=actions)
+    return MassenergizeResponse(actions)
```
BradHN1, line 227: Why this change?  This is not done consistently through the API.

src/_main_/utils/common.py
```diff
@@ -109,6 +109,10 @@ def rename_fields(args, pairs):
 def serialize_all(data, full=False):
   if not data:
     return []
+
+  if isinstance(data[0], dict):
```
Opoku-Agyemang, line 113: why are you doing data[0] ? what if its an int?

Opoku-Agyemang (approved):

Opoku-Agyemang (changes requested):
please address changes

src/api/utils/filter_functions.py
BradHN1: @Opoku-Agyemang - what is the best way to review this?

BradHN1: I tested this for Actions testing locally but fetching from Dev.  The additional time for applying the filter was 0.2 ms, which is small.  Prod is a much larger DB, so if it is 1000 times larger it may add 0.2 seconds.  We'll need to test this on Prod.

src/api/store/testimonial.py
```diff
@@ -26,6 +26,8 @@ def get_testimonial_info(self,  context: Context, args) -> Tuple[dict, MassEnerg
       
   def list_testimonials(self, context: Context, args) -> Tuple[list, MassEnergizeAPIError]:
     try:
+      filter_params = get_testimonials_filter_params(context.get_params())
```
BradHN1, line 29: The testiimonials.list route is only used by the community portal. 
I don't think it should use the filter_params, since none will be passed in.

src/api/store/action.py
```diff
@@ -34,7 +34,9 @@ def get_action_info(self, context: Context, args) -> Tuple[dict, MassEnergizeAPI
       return None, CustomMassenergizeError(e)
 
   def list_actions(self, context: Context, args) -> Tuple[list, MassEnergizeAPIError]:
-    try: 
+    try:
+      
+      filter_params = get_actions_filter_params(context.get_params())
```
BradHN1, line 39: I don't think we want the filter_params for the actions.list route.  Only used in community portal

src/api/services/userprofile.py
```diff
@@ -239,14 +242,15 @@ def list_users_for_community_admin(self, context, args) -> Tuple[list, MassEnerg
...
+    sorted = sort_items(users, context.get_params())
+    return paginate(sorted, context.get_pagination_data()), None
 
 
   def list_users_for_super_admin(self, context,args) -> Tuple[list, MassEnergizeAPIError]:
     users, err = self.store.list_users_for_super_admin(context,args)
     if err:
       return None, err
-    return serialize_all(users), None
+    return paginate(users, context.get_pagination_data()), None
```
BradHN1, line 253: users should be sorted for super admin list

BradHN1 (approved):
Great work - deploying to dev/canary

BradHN1:
Merging without final comments from Sam - will fix them later

PR #625, opened 2023-01-03
624 - Nudge should go weekly if admins haven't set comm prefs

BradHN1:
Fix the cadmin nudge which didn't go out to anyone.


abdullai-t (approved):
LGTM✅

PR #631, opened 2023-01-24
Show admins what to do

frimpongopoku:
## Works with this admin portal PR : https://github.com/massenergize/frontend-admin/pull/824

So the only new things that have happened here. 

- [X] Created a new route that creates the summary of what needs to be taken care of by the admin `/summary.next.steps.forAdmins`

And I just modified a few routes to do targeted listing. 
E.g users.listForCommunityAdmin used to list only users related to signed in admin's communities. 
I modified the route to also fetch users that match a list of given IDS. So now the route still behaves normally. But when you hit the same route and provide `user_emails` list, it will only retrieve the users that have emails in the given list, and ignore it's normal behaviour.

Did this in the following places: 

- [X] List All Users 
- [X] List admin messages 
- [X] List teams 
- [X] List testimonials 

This logic helps in retrieving specific items that the admin portal needs in preselecting table items, in case they are not available locally.


BradHN1 (changes requested):
Can you please add unit tests for the new route and the changes made to the other routes?

frimpongopoku:
In the end, unit tests will follow as usual. This deployment is only meant
for dev so that everyone can have a feel, as discussed in the meeting isnt
it? I want to do the tests when all the features are done, and we have what
we need for prod.

This way we dont have to be changing tests so many times as well.

src/api/services/summary.py
```diff
@@ -1,27 +1,64 @@
...
+        teams = content.get("teams", [])
+
+        # testimonials = serialize_all(testimonials)
+        # messages = serialize_all(messages)
+        # users = serialize_all(users)
+        # teams = serialize_all(teams)
 
-  def __init__(self):
-    self.store =  SummaryStore()
+        content = {
```
BradHN1, line 37: Seems like you fill content in the Store layer method, and then don't check it but fill it again here.

PR #644, opened 2023-02-14
Show admins what todo 2

frimpongopoku:
## Works with Admin Frontend API : https://github.com/massenergize/frontend-admin/pull/845 

- [X] Introduced a new route `/summary.get.engagement` to summarise community engagements, and differentiated things admins need to from community engagements. It collects user sign in footages, number of actions completed, number of actions in user todolists for all users in an admin's communities, based on a given datetime range. 

See admin portal PR for Demo

### DIFF
--- a/src/_main_/utils/common.py
+++ b/src/_main_/utils/common.py
@@ -106,11 +106,14 @@ def rename_fields(args, pairs):
   return args
 
 
-def serialize_all(data, full=False):
+def serialize_all(data, full=False, **kwargs):
+  medium = (kwargs or {}).get("medium", False)
   if not data:
     return []
   if full:
     return [d.full_json() for d in data]
+  elif medium: 
+    return [d.medium_json() for d in data]
   return [d.simple_json() for d in data]
 
 
--- a/src/_main_/utils/footage/FootageConstants.py
+++ b/src/_main_/utils/footage/FootageConstants.py
@@ -103,7 +103,7 @@ class FootageConstants:
 
     @staticmethod
     def on_user_portal():
-        return PLATFORMS["ADMIN_FRONTEND_PORTAL"]["key"]
+        return PLATFORMS["USER_FRONTEND_PORTAL"]["key"]
 
     @staticmethod
     def create():
--- a/src/_main_/utils/footage/spy.py
+++ b/src/_main_/utils/footage/spy.py
@@ -291,11 +291,14 @@ class Spy:
                 else UserProfile.objects.filter(email=ctx.user_email).first()
             )
             act_type = kwargs.get("type", None)
+            portal = kwargs.get("portal", None)
+            portal = portal if portal else FootageConstants.on_admin_portal()
             footage = Spy.create_footage(
                 actor=actor,
                 activity_type=act_type,
                 by_super_admin=ctx.user_is_super_admin,
                 item_type=FootageConstants.ITEM_TYPES["AUTH"]["key"],
+                portal = portal
             )
             groups = actor.communityadmingroup_set.all()
             communities = [g.community for g in groups]
--- a/src/_main_/utils/footage/spy.py
+++ b/src/_main_/utils/footage/spy.py
@@ -349,7 +349,7 @@ class Spy:
         return (
             Footage.objects.filter(
                 portal=FootageConstants.on_admin_portal(), by_super_admin=True
-            )
+            ).distinct()
             # .exclude(actor=user)
             .order_by("-id")[:LIMIT]
         )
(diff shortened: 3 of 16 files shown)


src/_main_/utils/common.py
```diff
@@ -106,11 +106,14 @@ def rename_fields(args, pairs):
   return args
 
 
-def serialize_all(data, full=False):
+def serialize_all(data, full=False, **kwargs):
+  medium = (kwargs or {}).get("medium", False)
```
BradHN1, line 110: This seems kludgy.  Just so you can pass the filter flags back for the community.  It might make sense if there was a medium_json for all the models, but that is not the case.  Can you add the filter flags into community.simple_json which I think will be plenty fast given there aren't and will never be too many filter flags.

frimpongopoku: It isnt actually @BradHN1  🤣 . Let me share what I was thinking and after, if you still think it's "kludgy" we will find a better way 🤣 🤣. 

So there are two parts to this. 

Just like you, I initially thought that I could just add the flags to the simple_json function. I didnt do it because, the flags will now make "simple_json" very heavy. Not in terms of the number of items in the object it returns, but in terms of how long it takes to actually finish running the "simple_json". You notice that when we are trying to get the flags enabled for a community or a user, we actually do a DB fetch for all flags, then do a bunch of other comparisons in loops before we arrive at the flags that are activated for a particular community/user. These extra milliseconds **will** pile up when are calling long lists of communities/users (which we already do in so many places). All of a sudden, "fast light weight list of communities" wont be showing up as quickly anymore. Following this, the only other option I had,  was to call "full_json" during serialization for all the "communities.listFor...." routes, which would have also given us too many unnecessary information for reach community (that would also waste too much time when loading the whole list of communities).

And that is why I just introduced `medium_json`. Doing so, I had to modify our `serialize_all` function. Here is one thing to note, I didnt only modify the structure of the function to forcefully fit in "`medium_json`". You will see that the third argument, isnt specifically "`medium`" it is `**kwargs`. Which now gives us space to pass in anything we want, as many things as we want, and also allow us to modify the `serialize` function however we want, for any special cases that arise without having to track back to it's old uses and editing them to match the new structure. 
So in short, this is actually the last time we would have to change structure of `serialize_all`. Yet, we now have the ability to tweak anything in the function to suit any nuanced cases we meet in future. 

So you see with this new change, any calls to 
`serialize_all(objects,False)` is valid. 
`serialize_all(objects, False, medium=True)` is also valid -- and in future 
`serialize_all(objects, False, something_new=Flase, some_other_thing=[1,2,3], some_other_thing_again=True)` will also be valid.

And all this solves the problem of keeping `simple_json` light weight as intended,  keeping `full_json` for when we need everything, and still catering for the weird case of us needing only lightweight content for all communities, but with their feature flags.

BradHN1:
I still think it is kludgy, and would like it the other way.
An option would be to only return the feature flags in the case that an Admin was making the call, if you are worried about speed.  But I think the speed implications would be milliseconds, not noticable.

frimpongopoku:
Brad, it isnt kludgy, it really isnt. This implementation does not hinder
anything, it only adds to the structure we have.

BradHN1:
Please do it the way I suggested.  We can add timing to see if it really takes more than a couple ms, and improve on it later.

src/_main_/utils/common.py
```diff
@@ -106,11 +106,14 @@ def rename_fields(args, pairs):
   return args
 
 
-def serialize_all(data, full=False):
+def serialize_all(data, full=False, **kwargs):
```
Opoku-Agyemang, line 109: I'd much prefer the following.
Define an enum like this:

```
from enum import Enum

class SerializationType(Enum):
    SIMPLE = "s"
    FULL = "f"
    MEDIUM = "m"
    # ... we can add as many as we want here
```

Then over here we can simplify this as 
```suggestion
def serialize_all(data, serialization_type: SerializationType= SerializationType.FULL):
```

Then you can go on and do any gymnastics in the models when we call json

Opoku-Agyemang (changes requested):

BradHN1 (approved):
LGTM
Will address comments on serialize_all

src/_main_/utils/common.py
BradHN1: For now I will deploy with the feature flags returned with simple_json.  this will not slow anything measurably.

src/_main_/utils/common.py
BradHN1: can make this change later.  For now feature flags returned in simple_json for communities

BradHN1 (approved):

PR #657, opened 2023-03-07
User Events Nudge

abdullai-t:
### Overview
This pull request adds a new feature to our API  to nudge users of a community of upcoming events i.e events created by the user's community, events shared with the community and open events based on the user's communication preferences.

### DB changes
`published_at` is now added to the `Event` table to track when an event went live. This will help us get all events that went live between nudge cycles.

### How users and events are collected

- Get all communities that are live, not archived and not blocked by the communication pref feature flag
- For each of the communities, get all events that are
     1. open
     2. shared with the community
     3. created by the community
     4. published and the published_at date falls between the two nudge cycles
-  for each of the communities, get  all users(exclude cadmins, sadmins and vendors) 
    1. get the user's preferred notification frequency and get all events that went live within that period.
    2. Users will only receive a nudge if the last nudge date + notification frequency from the user pref matches the current date.

### Other Changes
The `/downloads.sample.user_report` endpoint is added to allow cadmins and sadmins to have a sample of the user events nudge sent to their email.


src/api/store/summary.py
```diff
@@ -249,18 +249,18 @@ def next_steps_for_super_admins(self, context: Context, args):
             )
 
         testimonials = Testimonial.objects.values_list("id", flat=True).filter(
-            is_approved=False, is_deleted=False
+            is_deleted=False
```
BradHN1, line 252: Why these changes to summary.py?  These are not correct changes.

BradHN1 (approved):
LGTM - for testing in Dev

PR #658, opened 2023-03-08
MOU Signing For Admins

frimpongopoku:
## This PR works with frontend PR : https://github.com/massenergize/frontend-admin/pull/903

**[Find Feature Spec Here](https://docs.google.com/document/d/1hggx_5QWiZ0Avui-4AMj9uykjvmkd8PHjHz_TBGR1Ls/edit)**

### New Table (PolicyAcceptanceRecords) 
Will keep records of community admins signing their Memorandum of Understanding (MOU) (as an example) every year. 
Each community admin can have multiple (As they sign yearly, they will have multiple records).

### New Task 
A task will be created to run a function everyday at e.g. 8AM. Based on the requirements on this [sheet](https://docs.google.com/document/d/1hggx_5QWiZ0Avui-4AMj9uykjvmkd8PHjHz_TBGR1Ls/edit), this is what the function will do. 

1. It will comb the community admin table (communityAdminGroup)
2. For every admin, we check if its been a year since they last signed the MOU.
3. If it has been a year>, check if  they have been sent an email already. If Not, send!
4. If the admin has been sent an email already, check the timestamp for the last time an email was sent, and calculate if it's been more than a month ( Specified in the requirement sheet)
5. If it hasnt been a month, ignore! Otherwise, send an email again, and record the new timestamp.

### Table definitions 
`user` :  A relationship to the userProfile object 
`signed_at` :  Timestamp when user signs the MOU  (DateTime Field)
`last_notified` : An array of timestamp records for when notification emails are sent ( JSON Field (array))
`type` :  Text identifier that specifies what the record is about (Default = "MOU") 
_This way helps so we can use the same table for other scenarios that users need to sign/accept something_

### Tests 
- [X] I have added tests for the `send_admin_mou_notification` 
Test for route will follow up soon
- [X] Test for accepting/declining MOU route

### New Route 
`users.mou.accept`: Thats the route admins will hit when they choose "accept/deny" from the  admin frontend portal 

### New Installations ⚠️ 
You need to install `xhtml2pdf` if you dont already have it.

### New Migrations ⚠️  
You would have to regenerate migrations, and migrate. There is a new table as stated above, and I also added `key` to policies(crucial). 

## How To Get This To Work

1. I have added the `send_admin_mou_notification` to the `FUNCTIONS` object in `jobs.py`. So When this is deployed to (Dev,Canary, Prod) a super admin (@BradHN1 ) needs to create a task, and schedule the MOU function to run everyday at a certain time!.
2. When we are ready to deploy to PROD, email templates on postmark need to pushed to the production server and the template_id needs to be swapped.


src/database/models.py
```diff
@@ -957,6 +969,35 @@ class Meta:
         db_table = "user_profiles"
         ordering = ("-created_at",)
 
+class PolicyAcceptanceRecords(models.Model):
```
BradHN1, line 972: @frimpongopoku - there is a model named Policy, which I think we should use to keep the policy which the Admin is agreeing to.  This model should have a link to the version of the policy they agree to.

abdullai-t:
I think you forgot to add `xhtml2pdf` to the requirements. Please add that and fix the failing tests

BradHN1 (approved):
LGTM
remade migrations and some clean-up

PR #675, opened 2023-04-13
Admins can now see user visit logs

frimpongopoku:
## Works with Admin Portal PR:  https://github.com/massenergize/frontend-admin/pull/925

- [X] Just added a route that retrieves a specific user's sign in records on the user portal.


BradHN1:
@frimpongopoku - I think this should probably be done with the  UserProfile visit_log, not the footages.  For one thing - we only need to report the dates that a user visited the portal, not the times which is too much info and not the cadmin's business.  Second - since we already record when users visit the portal on the visit_log (including anonymous users), I don't think we should be recording user signin to the community portal in Footages. That may be saving too much data, and from looking at it there are generally multiple records per user signing in on a given day.  Too much data.  I think the footage should track what admins do on the admin portal.

frimpongopoku:
There shouldnt be multiple user records of signins per day. Unless a user just enjoys signing out and signing back in everytime its only going to be one sign in per day. 

Also, I was under the impression, saving the sign in logs on the user object in the json area was just a quick fix we needed at the time 🤔. Moving the logging into the footages table is a better approach.

The way we are collecting logs as an array on the user object is not ideal (it only works best for guest users). Putting everything in a proper DB table allows us to conduct more complex and very specific searches

frimpongopoku:
If we record multiple sign-ins of the same user from different browsers, that just means everything is working as it should.  And that would still happen independent of how we decide to store the data. Tracking our records in a table, is the best foundation we would want to lay. But, if we are worried about cost then sure, no problem. For user portal sign-ins, I will work with the items on the user object.

frimpongopoku:
Also, another thing to note is that if we stop recording user sign-ins on the portal in the table, user engagement will also take  hit. 
To use the visit_log on the user object for that, I would have to find every user in a community first, then manually compare the dates and count which sign in logs are in the specified time range.....  

Should I go ahead? @BradHN1

BradHN1 (approved):
LGTM
It's ok as long as we don't report more information than necessary for the Admin.

PR #681, opened 2023-04-28
Notify with sockets

frimpongopoku:
### What happened here? 

- [X] I installed `channels 3.0.5` (its available in requirements.txt) NB: It needs to be version 3.0.5. 
- [X] Switched our app from `wsgi` app to `asgi` app. The changes are in the `settings.py, asgi.py` files i checked in. 
- [X] Created a new app `socket_notifications`. Thats where we can handle all socket routes and consumer classes we need for now and in future. 
- [X] Added one new socket route `ws/me-client/connect/`. This is the route the frontend must hit when the app is loading to setup the connection between the frontend and backend.  
- [X] I also updated `/auth.login` for admins to calculate and record the expiration time of the particular user in session. So that it can be available to our socket consumers to use to track.

### How to test 
Check inside `user_session_tracker_consumer.py` file. Inside the `track` function, there is a commented line that says "FOR TESTING". 
You can uncomment that and it will forcefully expire  your admin in 15 seconds (You can change this during testing if you want). Just so you can have a feel of the flow of everything between the F.E and the B.E when the events are actually triggered.


src/_main_/settings.py
```diff
@@ -40,7 +40,7 @@
 # Database selection, development DB unless one of these chosen
 IS_PROD = False
 IS_CANARY = False
-IS_LOCAL = False
+IS_LOCAL = True
```
abdullai-t, line 43: pls revert this

abdullai-t (approved):
LGTM ✅

frimpongopoku:
Actually I shld have added, my settings.py is behind on so many things and
doesnt have lots of credentials, so whatever u have should be whats
 everyone has in their environment should be the source of truth. Mine is
just there for you to inspect and see the changes related to the socket
stuff I just added.

src/_main_/settings.py
BradHN1: don't worry - I'll fix this

src/api/handlers/auth.py
```diff
@@ -47,6 +48,9 @@ def login(self, request):
     # if the signin is from an admin site then set it to 24 hrs
     if(context.is_admin_site):
       MAX_AGE = ONE_DAY
+      expiration_time = get_date_and_time_in_milliseconds(hours=24)
+      print("LETS SEEE EXPIRATION TIME", expiration_time)
```
BradHN1, line 52: should be if (DEBUG)

BradHN1 (approved):
LGTM  - This is really cool, I don't really understand how it works, wkll learn,

PR #688, opened 2023-05-03
now records mou footage

frimpongopoku:
- [X] Now records a footage when an admin signs, or denies the MOU, and it shows on the frontend portal like this 

Screenshot 2023-05-03 at 11 49 42


BradHN1 (approved):
LGTM

PR #694, opened 2023-05-08
Bug Fix: Postmark footage Error

abdullai-t:
####  Summary / Highlights
This PR fixes the issue where the footage actor is null when admins reply to user messages from their email app. The issue occurs because the user_id and user_email fields in the API's context object are null since the user is not logged in. To solve the problem, we pass and use the sender email from the reply to get the associated user and use that user as the actor for the footage. This approach solves the bug by properly associating the actor with the footage, even when the user is not logged in.

### DIFF
--- a/src/_main_/utils/constants.py
+++ b/src/_main_/utils/constants.py
@@ -2,6 +2,7 @@
 This file contains Global constants used throughout the codebase.
 """
 
+import os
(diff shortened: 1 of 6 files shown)


BradHN1 (approved):
LGTM

BradHN1 (approved):
LGTM

PR #697, opened 2023-05-12
api support for new user visits changes

frimpongopoku:
### Related to frontend PR: https://github.com/massenergize/frontend-admin/pull/961

- [X] Made changes to fallback on "visit_log" if there is nothing recorded by footages 
- [X] Made changes to load in a few (5) records in simple_json. Its used on the frontend to determine which users have more info....


BradHN1 (approved):
LGTM

PR #704, opened 2023-05-22
Sort communities by name on list routes

BradHN1:
Fix so admin lists of communities are sorted rather than a random order.


abdullai-t (approved):
LGTM

PR #708, opened 2023-05-24
Carbon calculator changes

BradHN1:
1.  Implement date dependent values for emission reduction average:
     Previously, carbon  reductions for actions came from a static value in the cc Action model:
             calculator_action.average_points. 
     These values now come from a method: 
             CarbonCalculator.AverageImpact(calculator_action, date_completed)

     to do this we have added the average values as Carbon Calculator default values in Airtable, to be imported into the database.  Example for the action 'weatherization', where the second value includes a validation date of 2022-09-01::
        weatherization_average_points	default	3281			        2023-05-15 20:38
        weatherization_average_points	default	3216     2022-09-01	2023-05-15 20:38

2.  Add is_deleted for carbon calc actions


frimpongopoku (approved):
In terms of code implementation, nothing immediately jumps at me so -- LGTM ✅ 

However, in regards to what is actually happening in our carbon calculator, I have no idea.  I dont know how that works.

That said, @BradHN1  maybe when there is time you could take the team through or  better, document what exactly happens in the carbon calculator aspect of the app? 

I would really like to understand the workings of what happens there 🤔...  and I'm sure the others and people to come would benefit from this as well.

Things like: 
Where do we get the carbon values from?  Do they change? 
How we are relating a user's actions to a carbon score...  and all that stuff. That would be helpful!

PR #711, opened 2023-05-26
Three fixes

BradHN1:
1. Fix the communities.listForCommunityAdmin (or super admin) to properly order the list;
2. Fix the ME_SUPPORT_TEAM_EMAIL
3. Fix and diagnostic print for the MOU notificaiton nudge


frimpongopoku (approved):
Just seen @BradHN1  yes, it actually doesnt cater for people who have never signed before 🤣 💀 . But yeah, all the other parts LGTM

PR #716, opened 2023-05-31
Action Users List Api Changes

abdullai-t:
#### Newly Added
- An download endpoint was added to generate and send action users list to admins via email
- `action_users` was also added to the `to_json` of the actions model


BradHN1 (approved):
LGTM

PR #722, opened 2023-06-12
Filter bad words from username

frimpongopoku:
### Works with frontend PR : https://github.com/massenergize/frontend-portal/pull/1208 

Just modified Mattea's username validation code a bit. 

- [x] There was a bug with username validation. Any username always returned taken and suggested a new one even though the username was unique. So I fixed that one.
- [x] I introduced a new plugin "World filter". That is the plugin that checks if the username contains any profane words. Here is the list of profaned words that the plugin uses https://github.com/dariusk/wordfilter/blob/master/lib/badwords.json . If you find that there is a bad word that is not in the list, you can add it too, like I did here: 
Screenshot 2023-06-12 at 12 41 08

### DIFF
--- a/src/api/handlers/userprofile.py
+++ b/src/api/handlers/userprofile.py
@@ -89,9 +89,9 @@ class UserHandler(RouteHandler):
         context: Context = request.context
         args: dict = context.args
 
-        data, err = self.service.validate_username(args["username"])
+        data, err = self.service.validate_username(args.get("username"))
         if err:
-            return err
+            return MassenergizeResponse(error=err)
         return MassenergizeResponse(data=data)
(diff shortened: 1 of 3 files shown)


abdullai-t (approved):
LGTM ✅

BradHN1 (approved):
LGTM

BradHN1:
I've deployed this and it needs more work.  This is the response I got when I tried a bad word:

The username 'FuckU' is taken, how about ''?

It should say the username is not allowed due to vulgarity.  It doesn't need to suggest a username ''.
The vulgarity rules should also be applied to the First and Last name

PR #731, opened 2023-06-21
New Route for Event Sharing

### LINKED ISSUE #730: Event sharing needs a separate events.share route
**Describe the bug**
the events.update route prevents non-authorized people from sharing events.  

Create a separate events.share route used from the Admin portal to share an event from another community to your community

Require either superadmin access or cadmin of the community that the event is being shared to.
Check that event can be shared to the community .

abdullai-t:
####  Related [issue](https://github.com/massenergize/api/issues/730)
closes #730 

#### Changes
- Added `/events.share` route for event sharing instead of `/events.update` which was causing issues for unauthorised users.
- Fixed broken tests
- Added unit tests for the new route

### DIFF
--- a/src/api/store/event.py
+++ b/src/api/store/event.py
@@ -896,3 +896,23 @@ class EventStore:
     except Exception as e:
       capture_message(str(e), level="error")
       return None, CustomMassenergizeError(e)
+    
+
+
+  def share_event(self, context: Context, args) -> Tuple[dict, MassEnergizeAPIError]:
+    try:
+      event_id = args.pop("event_id", None)
+      event = Event.objects.filter(id=event_id).first()
+      shared_to = args.pop("shared_to", None)
+      if shared_to:
+        first = shared_to[0]
+        if first == "reset": 
+          event.shared_to.clear()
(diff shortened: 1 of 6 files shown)


src/api/store/event.py
```diff
@@ -893,6 +893,26 @@ def rsvp_remove(self, context: Context, args) -> Tuple[dict, MassEnergizeAPIErro
...
+
+
+  def share_event(self, context: Context, args) -> Tuple[dict, MassEnergizeAPIError]:
+    try:
+      event_id = args.pop("event_id", None)
+      event = Event.objects.filter(id=event_id).first()
+      shared_to = args.pop("shared_to", None)
+      if shared_to:
+        first = shared_to[0]
+        if first == "reset":
```
BradHN1, line 909: There should be a way for a community admin to unshare just to their community or communities, not to all communities.  It looks like this would unshare to all communities which if a cadmin shouldn't be able to do.

BradHN1 (approved):
LGTM - will merge this so a cadmin can share events now.  Let's address the further unsharing issue next

PR #733, opened 2023-06-26
users will now be able to submit testimonials

frimpongopoku:
## Related ticket :  https://github.com/massenergize/frontend-admin/issues/1025#event-9630559477

- [X] submitting testimonial wont give the "unexpected argument user_email" anymore

## DEMO 

https://github.com/massenergize/api/assets/26961591/1d3df220-70a8-440f-a6b2-4d22a8128f80

### DIFF
--- a/src/api/store/testimonial.py
+++ b/src/api/store/testimonial.py
@@ -74,7 +74,7 @@ class TestimonialStore:
       action = args.pop('action', None)
       vendor = args.pop('vendor', None)
       community = args.pop('community', None)
-      user_email = context.user_email
+      user_email = args.pop("user_email") or context.user_email
 
       args["title"] = args.get("title", "Thank You")[:100]
 
--- a/src/api/store/testimonial.py
+++ b/src/api/store/testimonial.py
@@ -74,7 +74,9 @@ class TestimonialStore:
       action = args.pop('action', None)
       vendor = args.pop('vendor', None)
       community = args.pop('community', None)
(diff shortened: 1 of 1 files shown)


src/api/store/testimonial.py
```diff
@@ -74,7 +74,7 @@ def create_testimonial(self, context: Context, args) -> Tuple[dict, MassEnergize
       action = args.pop('action', None)
       vendor = args.pop('vendor', None)
       community = args.pop('community', None)
-      user_email = context.user_email
+      user_email = args.pop("user_email") or context.user_email
```
abdullai-t, line 77: For security reasons, we would want the backend to be the sole source of truth for user credentials (email and ID) for routes like this.

abdullai-t: We can use the context to get the user instead of passing an email or id from the frontend

frimpongopoku:
Yes, I saw that. But if we do that, user_email still stays in args! And
thats whats causing the issue. So at the end of the day we need to remove
it before updating. But, I also didnt want to remove it completely, cos
then we would have to investigate and find out really why we are passing it
in there. That would need to be a slow process and would require a lot of
testing for us to be sure…


This bug is in prod, so it needs to be fixed quickly. This is the quickest
way to fix, while making sure we are not messing with any other things.

abdullai-t (approved):

BradHN1 (approved):
LGTM - comments added

PR #737, opened 2023-06-28
users can now update testimonials with images

abdullai-t:
####  Related [ISSUE](https://github.com/massenergize/frontend-admin/issues/1024#event-9630558191)
Works with [PR](https://github.com/massenergize/frontend-portal/pull/1223)


src/api/handlers/testimonial.py
```diff
@@ -102,7 +104,12 @@ def submit(self, request):
 
     # user submitted testimonial, so notify the community admins
     user_submitted = True
-    testimonial_info, err = self.service.create_testimonial(context, args, user_submitted)
+    is_edit = args.pop("testimonial_id", None)
```
BradHN1, line 107: this does not look like it will work.  testimonials_update requires 'id' in args, which would be the same as 'testimonial_id'.

BradHN1:
Please check that the code for updating user submitted testimonials, actions, events and vendors all work the same way.  They look different now and they should be as similar as possible.

src/api/handlers/testimonial.py
abdullai-t: the `id` is added to `args` on the frontend.  I am passing testimonial_id as a way of telling the backend this request is an update request.

abdullai-t:
> Please check that the code for updating user submitted testimonials, actions, events and vendors all work the same way. They look different now and they should be as similar as possible.

@BradHN1, updating submitted contents works. Here is a  [Demo](https://github.com/massenergize/frontend-portal/pull/1223). The code for actions, events, and vendors is consistent on both the backend and frontend. The update code for a testimonial on the backend accepts `id`  instead of the usual `{item}_id (testimonial_id)`. _Changing it to be consistent with the others will require changes on the admin portal, frontend portal, and also api._  That is why I went with the flow of the existing code

src/api/handlers/testimonial.py
BradHN1: It looks like frontend-portal passes in id, and frontend-admin passes in testimonial_id.   In my opinion, both should pass in 'id' and if that integer corresponds to a testimonial that would be an update request

BradHN1 (approved):
LGTM

PR #738, opened 2023-06-28
Action users Improvement

abdullai-t:
#### Changes
- [x] Added carbon impact to action users object in actions full_json


BradHN1 (approved):
LGTM

PR #739, opened 2023-06-29
media lib footage resolved

abdullai-t:
#### Why Error and Fix
The` .set` method of the M2M relationship accepts a list of `ids,` but we were passing a list of objects.  This Pr fixes this issue by replacing the list of objects with the list of ids


BradHN1 (approved):
LGTM

PR #744, opened 2023-07-17
Events nudge

abdullai-t:
#### Changes
- [x] User and cadmin events nudge now checks the audience type in feature flag to either include or exclude communities or users.


BradHN1 (approved):
LGTM

PR #750, opened 2023-07-22
Email Verification Template Improved

abdullai-t:
#### Changes
- [x] Added a new route to handle the processing and sending of verification emails to users through postmark


BradHN1 (approved):
LGTM - test in dev

PR #754, opened 2023-07-27
Postmark Email Templates Management Follow Up

abdullai-t:
#### Change
- Moved template fetch from constants file.


BradHN1 (approved):
LGTM

PR #756, opened 2023-08-01
Fix Zipcode error

BradHN1:
Crash in dev was due to some bad data in the database, a location without a zipcode.  Added protection


ntdkhiem (approved):
LGTM!

abdullai-t (approved):

PR #764, opened 2023-08-09
Email-Password Signup Email Verification Now Uses Postmark

abdullai-t:
#### Changes
- [x] Added a new route to handle the processing and sending of verification emails to users through postmark


BradHN1 (approved):
LGTM

PR #784, opened 2023-08-25
Search and Filter by Community Fixed

abdullai-t:
#### Changes
- [x] Given that `UserProfile` data is not frequently updated, the search and filtering of users by community is now performed using the `CommunityMember` table.


BradHN1 (approved):
LGTM

PR #785, opened 2023-08-29
Events Location Improvement

abdullai-t:
#### Changes
- [x] Added building name and room number to the location object


src/_main_/utils/common.py
```diff
@@ -170,6 +172,8 @@ def extract_location(args):
     "state": args.pop("state", None),
     "zipcode": args.pop("zipcode", None),
     "country": args.pop("country", 'US'),
+    "building": args.pop("building", None),
+    "room": args.pop("room", None),
```
BradHN1, line 176: This is duplicate code, which is not used anywhere.
I suggest either using extract_location from within parse_location, or eliminating extract_location

BradHN1 (approved):
LGTM

BradHN1:
For clarity: this doesn't add the fields to the Location model (not used by Event) but to the JSON field.

PR #791, opened 2023-09-11
Fix: Event Nudge Pref Link 

abdullai-t:
#### works with
-  Portal [PR](https://github.com/massenergize/frontend-portal/pull/1265)
- Admin [PR](https://github.com/massenergize/frontend-admin/pull/1092)

#### Changes
- [x] Added login_method to the args of the login route
- [x] Updated the URL for changing communication preferences in event nudge to include user data and login method

### DIFF
--- a/src/task_queue/events_nudge/cadmin_events_nudge.py
+++ b/src/task_queue/events_nudge/cadmin_events_nudge.py
@@ -61,14 +61,15 @@ def get_comm_admins(com):
     admins = []
     user_list = []
 
-    all_community_admins = CommunityAdminGroup.objects.filter(community=com).values_list('members__preferences', "members__email", "members__full_name", "members__notification_dates")
+    all_community_admins = CommunityAdminGroup.objects.filter(community=com).values_list('members__preferences', "members__email", "members__full_name", "members__notification_dates", "members__user_info")
 
     for i in list(all_community_admins):
         admins.append({
             "pref": i[0] if is_viable(i) else default_pref,
             "email": i[1],
             "name": i[2],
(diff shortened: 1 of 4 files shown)


src/task_queue/events_nudge/cadmin_events_nudge.py
```diff
@@ -205,7 +213,8 @@ def send_events_nudge():
 
 def send_events_report(name, email, event_list):
     try:
-        change_preference_link = ADMIN_URL_ROOT+"/admin/profile/preferences"
+        cred = encode_data({"email": email})
```
BradHN1, line 216: Can you move this duplicated code to a utility routine for example add_credentials()?  I think we will be using it a number of links.

src/task_queue/events_nudge/cadmin_events_nudge.py
```diff
@@ -205,7 +205,8 @@ def send_events_nudge():
 
 def send_events_report(name, email, event_list):
     try:
-        change_preference_link = ADMIN_URL_ROOT+"/admin/profile/preferences"
+        cred = encode_data_for_URL({"email": email})
```
BradHN1, line 208: should this also include the login_method (or could that login_method always be added if an email is specified)?

abdullai-t: the login_method can be included.

BradHN1: If you have the utility routine add that may simplify the usage

BradHN1 (approved):
LGTM

Issue #798, opened 2023-09-16
Carbon Calculator data needs to be updated when source data changes


BradHN1:
**Describe the bug**
Recent change updated data within the carbon calculator, which needs to be updated in any database which uses it.

**To Reproduce**
Tahiru received the following error running locally, which will happen to anyone who didn't manually reload the data:
Exception: Carbon Calculator error: value for "air_source_hp_average_points" not found in CalcDefault

**Expected behavior**
Whenever the API Django project is started, if the calculator data csv files have been updated, these data should be reloaded into the database, avoiding the error from a missing variable.  Also if the carbon calculator version is updated, reload the data because the data models may have changed.

PR #799, opened 2023-09-16
Carbon Calculator data update

BradHN1:
Relating to issue #798

Force carbon calculator data load when either calculator version changes or data files are updated
New data model Version keeps track of calculator version and update time.


src/carbon_calculator/carbonCalculator.py
```diff
@@ -26,6 +28,47 @@
...
+    return datestamp
+
+def versionCheck():
+    version = Version.objects.all()
+    if not version:
+        version = Version.objects.create(version=CALCULATOR_VERSION, note="Initial Version")
+        print(version.note)
+        return False    # reload data
+    else:
+        version = version.first()
```
abdullai-t, line 52: If we have multiple versions. This code wouldn't return the latest

abdullai-t (approved):
LGTM✅

src/carbon_calculator/carbonCalculator.py
BradHN1: There shouldn't be multiple versions, but in case there are I changed it to .latest('id')

PR #801, opened 2023-09-21
Task: Content Spacing Correction

abdullai-t:
####  Summary / Highlights
A few deployments ago, we found a way to resolve the inconsistent spacing in our content between the text on the admin portal and how it displays on the community portal. However, there are still many pieces of content in the database that continue to use the old fix (adding double spacing in the richText field on the admin portal). In this PR, a task has been created to generates a report and fix contents in the db that needs fixing.

#### Testing Steps (Provide details on how your changes can be tested)
1. Open the admin portal.
3. create a task and select  `Process Content Spacing` as the automatic process.
4. open your email for the report generated.
5. Visit the community portal to observe the spacing correction in the descriptions of actions, events, and testimonials.


src/task_queue/jobs.py
```diff
@@ -2,6 +2,7 @@
 from task_queue.events_nudge.cadmin_events_nudge import send_events_nudge
 from task_queue.events_nudge.user_event_nudge import prepare_user_events_nudge
 from task_queue.nudges.postmark_sender_signature import collect_and_create_signatures
+from task_queue.nudges.contents_spacing_correction import correct_contents_spacing
```
BradHN1, line 5: comment on organization: this isn't a nudge.  Maybe put it in database_tasks.  And move the events_nudges into nudges

src/task_queue/nudges/contents_spacing_correction.py
```diff
@@ -0,0 +1,44 @@
...
+
+def correct_spacing_in_testimonials():
+    testimonials = Testimonial.objects.filter(is_deleted=False)
+    for testimonial in testimonials:
+        body = testimonial.body.replace("<p><br></p>", "").replace("<p>.</p>", "").replace("<p>&nbsp;</p>", "")
+        testimonial.body = body
+        testimonial.save()
+    return testimonials.count()
+
+def correct_contents_spacing():
```
BradHN1, line 33: Add a feature flag update-html-format, so we can update first one community then others.
Also we want to be able to run and get a report of what content in what communities would be updated first, before actually updating the content.  That could use a separate feature flag check-html-format

BradHN1:
There are numerous other instances of HTML text.  Would it be possible to pass a model and have it check the fields of type TextField, note whether changes were made and then if changes were made do the save()?  Some of those aren't HTML fields but it wouldn't hurt to do the check.

src/task_queue/nudges/contents_spacing_correction.py
BradHN1: @abdullai-t - what I meant is to have one task, which always will send the report, but if the feature flag is set, it will do the updates for that communities content.

BradHN1:
Couple of minor things:

For teams, the responsibility community is in the field primary_community.
In the report, if no community is found (for any item), leave it blank instead of N/A

BradHN1:
When you added the 'task' argument: res = func(task)
it broke all the tasks except the one which expected that argument.

BradHN1:
Tried testing with my local database. First I ran the report, second I ran the fix.. Both with all communities selected.  Third I ran the report again, and the report had the same number of instances.  So I think the data wasn[t fixed.  Did you test it this way?

src/task_queue/database_tasks/contents_spacing_correction.py
```diff
@@ -0,0 +1,133 @@
...
+            if hasattr(instance, "community"):
+                if instance.community in enabled_communities:
+                    field_value = getattr(instance, field_name)
+                    if field_value:
+                        count = len(re.findall(PATTERN, field_value))
+                        if count > 0:
+                            setattr(instance, field_name, field_value.replace("<p><br></p>", "").replace("<p>.</p>", "").replace("<p>&nbsp;</p>", ""))
+                            instance.save()
+            else:
+                setattr(instance, field_name, field_value.replace("<p><br></p>", "").replace("<p>.</p>", "").replace("<p>&nbsp;</p>", ""))
```
BradHN1, line 81: Under the second condition, field_value is not filled.

src/task_queue/database_tasks/contents_spacing_correction.py
```diff
@@ -0,0 +1,133 @@
...
+    try:
+        model, model_instances = get_model_instances(model_name, app_label)
+        for instance in model_instances:
+            if hasattr(instance, "community"):
+                if instance.community in enabled_communities:
+                    field_value = getattr(instance, field_name)
+                    if field_value:
+                        count = len(re.findall(PATTERN, field_value))
+                        if count > 0:
+                            setattr(instance, field_name, field_value.replace("<p><br></p>", "").replace("<p>.</p>", "").replace("<p>&nbsp;</p>", ""))
```
BradHN1, line 78: You have these patterns defined in PATTERNS, but then they are hardcoded here.  Could you loop over the patterns and do the replace that way?

abdullai-t:
@BradHN1 , I have address all the feedbacks.

src/task_queue/database_tasks/contents_spacing_correction.py
```diff
@@ -0,0 +1,110 @@
...
+     if hasattr(instance, "community") and instance.community in enabled_communities:
+         return True
+     elif hasattr(instance, "primary_community") and instance.primary_community in enabled_communities:
+         return True
+     else:
+         return False
+
+def process_spacing_data(task=None):
+    try:
+        communities = db_models.Community.objects.filter(is_published=True, is_deleted=False)
```
BradHN1, line 67: minor change - don't require is_published=True
For the few communities which are working on getting content published, we want to edit that as well

src/task_queue/database_tasks/contents_spacing_correction.py
```diff
@@ -0,0 +1,110 @@
...
+        if not flag or not flag.enabled():
+            return
+        enabled_communities = flag.enabled_communities(communities)
+
+        data = []
+        models = apps.get_models()
+        for model in models:
+            app_label = model._meta.app_label
+            for field in model._meta.fields:
+                if field.__class__.__name__ == "TextField" and app_label in ["database", "task_queue"]:
```
BradHN1, line 78: question - why include the models in task_queue?

src/task_queue/database_tasks/contents_spacing_correction.py
```diff
@@ -0,0 +1,110 @@
...
+                             
+        if len(data) > 0:
+            report =  write_to_csv(data)
+            temp_data = {'data_type': "Content Spacing", "name":task.creator.full_name if task.creator else "admin"}
+            file_name = "Content-Spacing-Report-{}.csv".format(datetime.datetime.now().strftime("%Y-%m-%d"))
+            send_massenergize_email_with_attachments(DATA_DOWNLOAD_TEMPLATE,temp_data,[task.creator.email], report, file_name)
+    
+        return True
+    except Exception as e:
+        capture_message(str(e), level="error")
```
BradHN1, line 106: if is_dev or running locally, print(str(e))

BradHN1:
Have you tested this with your local database, and when you update the spacing and run the report again, does it show no occurances of the strings being replaced?

abdullai-t:
> and when you update the spacing and run the report ag

Yes. it doesn't show already replaced content in the report

src/task_queue/database_tasks/contents_spacing_correction.py
```diff
@@ -0,0 +1,110 @@
...
+    writer.writeheader()
+    for row in data:
+        writer.writerow(row)
+    return response.content
+def get_community(instance):
+    if instance and hasattr(instance, "community"):
+        return instance.community.name if instance.community else ""
+    elif instance and hasattr(instance, "primary_community"):
+        return instance.primary_community.name if instance.primary_community else ""
+
```
BradHN1, line 40: else return ""
in case the instance doesn't have a community

src/task_queue/database_tasks/contents_spacing_correction.py
```diff
@@ -0,0 +1,110 @@
...
+                    field_name = field.name
+                    model_instances = get_model_instances(model_name, app_label)
+                    for instance in model_instances:
+                         field_value = getattr(instance, field_name)
+                         if field_value:
+                            count = len(re.findall(PATTERN, field_value))
+                            if count > 0:
+                                if is_feature_enabled(instance):
+                                    auto_correct_spacing(instance, field_name, field_value)
+                                else:
```
BradHN1, line 88: no need for the "else"
Have it always append the data for what is being changed, even if the feature is enabled

src/task_queue/database_tasks/contents_spacing_correction.py
```diff
@@ -0,0 +1,110 @@
...
+from sentry_sdk import capture_message
+from _main_.utils.emailer.send_email import send_massenergize_email_with_attachments
+from api.utils.constants import DATA_DOWNLOAD_TEMPLATE
+from database import models as db_models
+from django.http import HttpResponse
+import re
+
+
+FEATURE_FLAG_KEY = "update-html-format-feature-flag"
+PATTERNS = ["<p><br></p>", "<p>.</p>", "<p>&nbsp;</p>"]
```
BradHN1, line 13: One more pattern to correct: "<br />."

src/task_queue/database_tasks/contents_spacing_correction.py
```diff
@@ -0,0 +1,110 @@
...
+    flag = db_models.FeatureFlag.objects.filter(key=FEATURE_FLAG_KEY).first()
+    if not flag or not flag.enabled():
+        return False
+    enabled_communities = flag.enabled_communities(communities)
+    if hasattr(instance, "community") and instance.community in enabled_communities:
+         return True
+    elif hasattr(instance, "primary_community") and instance.primary_community in enabled_communities:
+         return True
+    else:
+         return False
```
BradHN1, line 68: If the model has a "community" or "primary_community" field, but that is not filled, and the feature flag is enabled, it should make the change.  I can update the routine as follows:
`    if hasattr(instance, "community"):
        if not instance.community or instance.community in enabled_communities:
            return True
    elif hasattr(instance, "primary_community"):
        if not instance.primary_community or instance.primary_community in enabled_communities:
            return True
    return False
`

src/task_queue/database_tasks/contents_spacing_correction.py
BradHN1: Fixed and pushed

src/task_queue/database_tasks/contents_spacing_correction.py
BradHN1: fixed

BradHN1 (approved):
LGTM

PR #803, opened 2023-10-02
User portal copyright with images

frimpongopoku:
## Related to frontend PR: 
Here is what has changed:
- [X] Creating  & Editing user content from the community portal now includes copyright and underage consent data.

### DIFF
--- a/src/api/handlers/action.py
+++ b/src/api/handlers/action.py
@@ -90,6 +90,19 @@ class ActionHandler(RouteHandler):
       .expect("image", "file", is_required=False, options={"is_logo": True})
       .expect("vendors", list, is_required=False)
       .expect("action_id", str, is_required=False)
+      .expect(
+            "underAge", bool,
+        ).expect(
+            "copyright", bool,
+        ).expect(
+            "copyright_att", str,
+        ).expect(
+            "guardian_info", str,
+        ).expect(
+            "size", str
+        ).expect(
+            "size_text", str
+        )
     )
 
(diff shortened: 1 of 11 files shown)


src/api/handlers/action.py
```diff
@@ -90,6 +90,19 @@ def submit(self, request):
       .expect("image", "file", is_required=False, options={"is_logo": True})
       .expect("vendors", list, is_required=False)
       .expect("action_id", str, is_required=False)
+      .expect(
```
BradHN1, line 93: instead of all this code, can the front-end pass in 'media_info' as a json structure, which then goes into media_info?   No need to see guardian and copyright stuff as part of an action or a vendor or all the places we create content.

BradHN1:
As suggested above - pass in media_info from front end as a json structure, with flexibility to change it in the future without updating many code files. 
We also need this info (in UserMediaUpload) for community home page images, all page setting images, any image that gets uploaded from the Admin portal.

src/api/handlers/action.py
frimpongopoku: Thats is possible, its just NOT good. We want to be able to validate all the parts.

frimpongopoku:
> As suggested above - pass in media_info from front end as a json structure, with flexibility to change it in the future without updating many code files. We also need this info (in UserMediaUpload) for community home page images, all page setting images, any image that gets uploaded from the Admin portal.

@BradHN1  I am not sure I understand what you mean here... what "info"?

PR #811, opened 2023-10-11
More be tweaks for mlibrary from feedback

frimpongopoku:
### Works with frontend PR: https://github.com/massenergize/frontend-admin/pull/1104

⚠️  Migrations 
- Added "publicity" field to  UserMediaUploads. Its to help us identify images that are marked as "public"  on the entire platform as suggested in the doc. 

- [X] Added new filter params to fetch all images that are marked as public. 
- [X] Media library search results now include "total" number info which is used on the media library footer.
- [X] Wrapped all request params needed for media stuff into a function so it can be used anywhere easily. (Will use the updates here to clean out what was done in the other PR)


src/api/store/media_library.py
```diff
@@ -196,7 +204,9 @@ def get_most_recent(self, args, context: Context):
             else:
                 query |= qObj
 
+        count = 0
         if not upper_limit and not lower_limit:
+            count = Media.objects.filter(query).distinct().count()
```
BradHN1, line 209: change to count = images.count() two lines later?  Seems this might avoid some DB access

src/api/store/media_library.py
```diff
@@ -196,7 +204,9 @@ def get_most_recent(self, args, context: Context):
...
             images = Media.objects.filter(query).distinct().order_by("-id")[:limit]
         else:
             images = (
```
BradHN1, line 212: if you go this path, count=0.  Is that the intention?

src/api/store/media_library.py
```diff
@@ -205,7 +215,27 @@ def get_most_recent(self, args, context: Context):
...
             )
-        return images, None
+        return images, {"total": count}, None
+
+    def get_public_images(self, args):
+        upper_limit = args.get("upper_limit")
+        lower_limit = args.get("lower_limit")
+        count = 0
+        if not upper_limit and not lower_limit:
+            count = Media.objects.filter(
```
BradHN1, line 225: same comment as above

src/api/store/media_library.py
```diff
@@ -255,7 +289,9 @@ def get_by_keywords(self, args):
             else:
                 query |= queryObj
 
+        count = 0
         if not upper_limit and not lower_limit:
+            count = Media.objects.filter(query).distinct().count()
```
BradHN1, line 294: same comments as above

src/database/utils/settings/model_constants/user_media_uploads.py
```diff
@@ -0,0 +1,7 @@
+
+
+from database.utils.settings.model_constants.events import EventConstants
+
+
+class UserMediaConstants(EventConstants):
+    pass
```
BradHN1, line 7: Is this completed?

src/database/models.py
```diff
@@ -1182,13 +1197,19 @@ class UserMediaUpload(models.Model):
     info = models.JSONField(null=True, blank=True)
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
+    publicity = models.CharField(
```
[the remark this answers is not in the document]
frimpongopoku: Noted

src/api/store/media_library.py
frimpongopoku: Yes @BradHN1, count is  0 every time after the first time we retrieve the actual count within the "if not upper and not lower_limit" block (Which essentially means every first time we run the query). I dont bother to recalculate the  count when the limit values are available because I know that by the time upper_limit and lower_limits are available, the user has definitely used the "load more" feature so it's the same query, nothing has changed and chances are the total count is the same (The frontend knows not to update the the total count value on 'load more' requests).

At the time I was saving us unnecessary database access, but now that I am explaining it to you, I think I can see how it is very possible that content would have changed either by (adding/deleting) by other admins at any given time, so its better to retrieve the updated count value as much as we can 🤣 🤣 🤣 🤣 🤣

src/database/utils/settings/model_constants/user_media_uploads.py
frimpongopoku: Yes @BradHN1, I currently need the exact methods and properties that EventConstants has, so I am extending from that class. This way, UserMediaConstants inherits all the properties of EventConstants. UserMediaConstants does not need anything special yet, that is why in itself, it has nothing defined. So yes, this is actually complete. (Its Object Oriented Programming)

BradHN1:
@frimpongopoku - is this superceded by PR 813, or additioinal?

frimpongopoku:
Yes I merged the things here into #813 so yeah, u can  close this one.

BradHN1 (approved):
LGTM - test in Dev

PR #820, opened 2023-10-25
new route to help img cropping for already uploaded image (to skip CO…

frimpongopoku:
### Works with frontend PR: https://github.com/massenergize/frontend-admin/pull/1108

- [X] Created a new route that reads the content of an image in our S3 bucket and returns the base64 string of the image. Which is need for cropping on the F.E. 

### Note 
To be able to implement cropping, we need to be able to read the image data. Doing that with a just-selected image from a user's device is straight forward. But for an image that has been uploaded, you need to be able to load the image via the link and try to read it's content. Doing that with the image links we have  gives a cross-origin error (CORS). That is why I had to make a route that reads the image data and sends it to the frontend to be able to continue. 

Things would have been easier and more straight forward if we could simply read the data from the image that is loaded. And I think that could probably be done with adding some new policies to our image buckets on AWS to looses the grip on accepted origins, but that might open us up to our images being used in various places we might not like... so as you can see the PR, I went with the B.E route which works nicely. 🤣 🤣 👌🏾


src/api/store/media_library.py
```diff
@@ -21,6 +22,16 @@ class MediaLibraryStore:
     def __init__(self):
         self.name = "MediaLibrary Store/DB"
 
+    def read_image(self, args):
+        try:
+            id = args.get("media_id", None)
+            image = Media.objects.filter(pk=id).get()
```
abdullai-t, line 28: I am not sure this query will work. try `Media.objects.filter(pk=id).first()`

abdullai-t (approved):
LGTM

BradHN1:
Unit tests for new route?

src/api/store/media_library.py
frimpongopoku: You do realise I did the test the whole thing before I made the PR right? 🤣 🤣 🤣. 
It works, try it locally yourself.

src/api/tests/common.py
```diff
@@ -31,7 +32,7 @@
 from django.core.files.uploadedfile import InMemoryUploadedFile
 
 RESET = "reset"
-
+ME_DEFAULT_TEST_IMAGE = "https://www.whitehouse.gov/wp-content/uploads/2021/04/P20210303AS-1901-cropped.jpg"
```
BradHN1, line 35: you know this image is a joke, because Sam thinks I look like Joe Biden

BradHN1 (approved):
LGTM
I see a bunch of media library tests have been commented out.  Do theuy need replaciing

src/api/tests/common.py
frimpongopoku: 🤣 🤣 🤣 🤣 okay great. I thought Biden was the core team's favorite guy 🤣 🤣. Then we will change it to something more light weight so that tests run faster.

frimpongopoku:
> LGTM I see a bunch of media library tests have been commented out. Do they need replacing

They have already been replaced in new PRs. I will clear them out soon.

BradHN1 (approved):
LGTM - test in dev

PR #837, opened 2023-11-09
get summary without feature flag

frimpongopoku:
- [X] If there are no feature flags, it will summarise duplicates for all communities

### DIFF
--- a/src/task_queue/database_tasks/media_library_cleanup.py
+++ b/src/task_queue/database_tasks/media_library_cleanup.py
@@ -11,14 +11,14 @@ from api.store.common import (
 )
 from api.store.media_library import MediaLibraryStore as media_store
 from api.utils.constants import MEDIA_LIBRARY_CLEANUP_TEMPLATE
-from database.models import FeatureFlag
+from database.models import Community, FeatureFlag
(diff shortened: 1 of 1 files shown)


src/task_queue/database_tasks/media_library_cleanup.py
```diff
@@ -33,7 +33,10 @@ def remove_duplicate_images(task):
              print("Generating hashes")
              result = media_store.generate_hashes(None, None, None)
 
-        communities = flag.enabled_communities()
+        if flag and do_updates: 
+            communities = flag.enabled_communities()
```
BradHN1, line 37: I think if the flag is enabled, but no communities selected, this will produce an empty queryset and it will not do any reporting on duplicates, but probably blow up.

BradHN1:
There is some test code that got in there, my mistake.  Note two calls to generate_hashes where I had added one before I saw you had.  Also an IF_LOCAL in the MEDIA_LIBRARY - doesnt' do anything except print out.

archx3:
@frimpongopoku @BradHN1 is this PR still useful?
I'd like to archive it if not.

PR #843, opened 2023-11-21
Broadcast Messaging V2

abdullai-t:
#### Works with[ Admin PR](https://github.com/massenergize/frontend-admin/pull/1114)

### DIFF
--- a/src/api/handlers/community.py
+++ b/src/api/handlers/community.py
@@ -36,7 +36,6 @@ class CommunityHandler(RouteHandler):
     self.add("/communities.others.listForCommunityAdmin", self.list_other_communities_for_cadmin)
     self.add("/communities.listForSuperAdmin", self.super_admin_list)
     self.add("/communities.adminsOf", self.fetch_admins_of)
-    self.add("/communities.listNoPagination", self.communities_list_no_pagination)
 
   def info(self, request):
     context: Context = request.context
@@ -308,16 +307,4 @@ class CommunityHandler(RouteHandler):
     if err:
       return err
     return MassenergizeResponse(data=community_completed_actions)
-
-
-  @admins_only
-  def communities_list_no_pagination(self, request):
-    context: Context = request.context
-    args: dict = context.args
-
-
-    communities, err = self.service.communities_list_no_pagination(context, args)
-    if err:
-      return err
-    return MassenergizeResponse(data=communities)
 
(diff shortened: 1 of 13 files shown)


src/api/handlers/community.py
```diff
@@ -36,6 +36,7 @@ def registerRoutes(self) -> None:
     self.add("/communities.others.listForCommunityAdmin", self.list_other_communities_for_cadmin)
     self.add("/communities.listForSuperAdmin", self.super_admin_list)
     self.add("/communities.adminsOf", self.fetch_admins_of)
+    self.add("/communities.listNoPagination", self.communities_list_no_pagination)
```
BradHN1, line 39: I don't think you we should need routes like this. There should be a URL parameter that sets the pagination length, which could be set to return all results in one query.

src/api/handlers/userprofile.py
```diff
@@ -36,28 +36,31 @@ def registerRoutes(self):
...
-      context: Context = request.context
-      args: dict = context.args
-      args, err = self.validator.expect("id", str, is_required=True).verify(context.args)
-      if err:
-        return err
-      visits, err = self.service.fetch_user_visits(context, args)
-      if err:
-        return err
-      return MassenergizeResponse(data=visits)
+        self.add("/users.listNoPagination", self.users_list_no_pagination)
```
BradHN1, line 49: same comment as above; I think a URL parameter should accomplish this, not a dedicated route.

src/api/handlers/community.py
abdullai-t: Thank you for the feedback

src/api/store/message.py
```diff
@@ -29,45 +30,23 @@ def get_schedule(schedule):
...
+        com = Community.objects.filter(id=id).first()
+        if com:
+            return com.logo.file.url if com.logo else None
+        
     
 
 
-def get_message_recipients(audience, audience_type, sub_audience_type, communities):
+def get_message_recipients(audience, audience_type):
     if not is_null(audience):
```
BradHN1, line 44: if is_null returns true, do you nee to have a "return None" statment?

src/api/store/message.py
```diff
@@ -383,16 +369,23 @@ def send_message(self, context, args) -> Tuple[dict, MassEnergizeAPIError]:
...
+            associated_community = None
             if not recipients:
                 return None, InvalidResourceError()
             
             user = get_user_from_context(context)
             if not user:
                 return None, InvalidResourceError()
             
+            if not context.user_is_super_admin:
+                associated_community = Community.objects.filter(id=communities[0]).first()
```
BradHN1, line 387: question - is there a case where multiple communities will be passed, perhaps for a community admin of multiple communities?  In that case, picking the first one won't always be correct.   (fix later)

BradHN1 (approved):
LGTM - test in dev

PR #886, opened 2024-01-22
Fix campaign handler validator and store issues

abdullai-t:
This pull request fixes issues in the campaign handler validator and store. It adds a missing "role" field in the validator and updates the store to return the correct objects.


archx3 (approved):
Great job mate.
I like the addition of messages to the exceptions

PR #928, opened 2024-02-23
Feat/User nudge cadmin selection

abdullai-t:
####  Summary / Highlights

This update introduces enhancements to the user event nudge feature, adding a new method to validate whether an event is eligible based on its nudge settings, and modifying the prepare_user_events_nudge method to follow a new task format, making it more resistant to errors.
Additionally, access to the task object is now enabled in the method to get community events. This should lead to a more powerful and resilient user event nudge feature.


archx3:
There's a conflict, can you resolve it?

abdullai-t:
@archx3, merge conflicts resolved

archx3 (approved):

archx3 (approved):

src/task_queue/nudges/user_event_nudge.py
```diff
@@ -90,42 +94,78 @@ def get_email_lists(users):
...
+        "EVERY_DAY": relativedelta(days=1)
+    }
+
+    if freq:
+        last_last_run = now - freq_to_delta.get(freq, relativedelta(days=0))
+
+    if settings.when_first_posted and event.published_at and last_last_run < event.published_at <= now:
+        return True
+    elif settings.within_30_days and event.start_date_and_time - now <= timezone.timedelta(days=30):
+        return True
```
BradHN1, line 131: The intention of "within_30_days" was that a user should be notified once about it when the event is about 30 days away.  As I read this, the user will be nudged every week when the event is less than 30 days away.

src/task_queue/nudges/user_event_nudge.py
```diff
@@ -90,42 +94,78 @@ def get_email_lists(users):
...
+
+    freq_to_delta = {
+        "EVERY_WEEK": relativedelta(weeks=1),
+        "bi-weekly": relativedelta(weeks=2),
+        "EVERY_MONTH": relativedelta(months=1),
+        "EVERY_DAY": relativedelta(days=1)
+    }
+
+    if freq:
+        last_last_run = now - freq_to_delta.get(freq, relativedelta(days=0))
```
BradHN1, line 126: i'm not sure whether this might be a problem.  The event nudge task generally runs daily, though users are generally nudged once a week based on their preferences.

src/task_queue/nudges/user_event_nudge.py
```diff
@@ -90,42 +94,78 @@ def get_email_lists(users):
...
+
+    if settings.when_first_posted and event.published_at and last_last_run < event.published_at <= now:
```
BradHN1, line 128: This looks like a bug, but it might just be confusing. 
How is "last_last_run < event.published_at <= now" evaluated?

src/task_queue/tasks.py
```diff
@@ -5,38 +5,48 @@
 from task_queue.helpers import is_time_to_run
 from .jobs import FUNCTIONS
 from api.services.utils import send_slack_message
-from _main_.settings import SLACK_SUPER_ADMINS_WEBHOOK_URL, IS_PROD
+from _main_.settings import SLACK_SUPER_ADMINS_WEBHOOK_URL, IS_PROD, IS_LOCAL, IS_CANARY
 
+ENV = "API Local" if IS_LOCAL else "API CANARY" if IS_CANARY else "API PROD" if IS_PROD else "API DEV"
```
BradHN1, line 10: unless things have changed, IS_LOCAL means the API is using the local database.  It can be running locally but pointing to a DEV or CANARY database.  Or has that been changed?

BradHN1 (commented):
Left some comments.

src/task_queue/tasks.py
archx3: Hi @BradHN1 I believe I answered this question last week, but for visibility, this new implementation is meant to prevent manually changing environments before deployments but reading them programmatically after they have been passed as env vars to our deployment commands.

We'll be adding the ability to point the local to another ENVs we want next week.

src/task_queue/nudges/user_event_nudge.py
abdullai-t: Yes, the current implementation will send the notification multiple times within 30 days. This issue will be addressed alongside the new direction outlined in @frimpongopoku's [Figma design](https://www.figma.com/file/broFyNtjZQINatcfJxShqA/Community-Profile-Rearrangment?type=design&node-id=359-2&mode=design).

src/task_queue/nudges/user_event_nudge.py
BradHN1: I understand this now.  I didn't realize you could do a<b<c meaning a < b and b<c.

src/task_queue/nudges/user_event_nudge.py
BradHN1: Here is what I think the problem is:

A user last received a notification three days ago, with once a week nudge preferences, and the nudger runs every day at noon.  The event was published last night.   
Today when the nudger runs at noon, last_last_run is yesterday at noon, and if settings.when_first_posted is true, the function is_event_eligible will return true.  But the user won't be nudged since it hasn't been a week since the last nudge.
Then tomorrow when the nudger runs again, the last_last_run will be after the event was published, so it won't be sent out.

abdullai-t: It's true that we are thinking about creating separate celery jobs for event-specific notifications. This means that when an admin adds a setting for an event, we can create a celery job that will send reminders to users about that event only, in addition to the general event nudge that goes out to users every week.

BradHN1:
I separate celery tasks mighjt violate the preferences which we have set up.  Make sure we don’t do that.  If a user says one email a week, we don’t want to send more than that.

abdullai-t:
We can implement this feature while respecting the user's communication preferences. With this out of the way, do we have the green light to proceed?

PR #1002, opened 2024-05-15
rc: dev to canay

archx3:


src/task_queue/models.py
```diff
@@ -55,10 +57,11 @@ class Task(models.Model):
         to=UserProfile
     )
     last_run = models.DateField(blank=True, null=True)
+    is_archived = models.BooleanField(default=False)
```
BradHN1, line 60: Do we want this model update?  Doesn't hurt anything but the decision was made not to use it.

src/api/store/community.py
```diff
@@ -1288,6 +1290,26 @@ def update_community_notification_settings(self, context, args) -> Tuple[dict, M
...
             
+            if activate_on:
+                more_info = notification_setting.more_info or {}
+                task_id = more_info.get("task_id")
+                if task_id:
+                    AsyncResult(task_id).revoke()
+                
+                eta = datetime.strptime(activate_on, "%Y-%m-%d") + timedelta(minutes=0)
+                
+                task_id = automatically_activate_nudge.apply_async(args=[notification_setting.id], eta=eta).id
```
BradHN1, line 1301: Question - queued task survive when a new version of API is deployed? I think so but want to be sure.

abdullai-t: I am not too sure about it but I think they do survive

src/task_queue/tasks.py
```diff
@@ -1,52 +1,71 @@
...
+
+def get_task_info(task):
+	if task:
+		return f"task: {task.name}, job: {task.job_name}, task_id: {task.id}"
+	
+def log_status(task, message):
+	task_name = get_task_info(task)
 
-ENV = "API Local" if IS_LOCAL else "API CANARY" if IS_CANARY else "API PROD" if IS_PROD else "API DEV"
+	if task.status == FAILED:
```
BradHN1, line 27: someones linter isn't going to like these tab indentations.   Better to set them right the first time.

src/Makefile
```diff
@@ -139,3 +139,7 @@ migrate-canary:
 migrate-local:
 	DJANGO_ENV=local python manage.py migrate
 .PHONY: migrate-local
+
+
+create-cache-table-local:
```
BradHN1, line 144: Is this used by anything?  if not should remove it.

src/_main_/settings.py
```diff
@@ -207,6 +207,13 @@
 #     }
 # }
 
+CACHES = {
```
BradHN1, line 210: is this used?  if not remove.

abdullai-t: George has some incoming changes that will use it

src/task_queue/models.py
abdullai-t: We might need it in the future, so let's keep it for now.

PR #1120, opened 2024-07-24
Add Translator utils

archx3:
####  Summary / Highlights
This PR introduces Translator utils and suite of enhancements to our translation capabilities. It includes the implementation of translation utilities with Google and Microsoft services, the introduction of a languages.json file for supported languages, and refined JSON translation management using regex patterns and a TextHashService. 

Addresses #1086

* [ ] I've given my PR a meaningful title.
* [ ] I linked this PR to the relevant issue.


src/_main_/utils/translation/__init__.py
```diff
@@ -1,19 +1,45 @@
...
+JSON_EXCLUDE_VALUES = {
+    "null",
+    "undefined",
+    "None",
+    "True",
+    "False",
+    "null",
+    "false",
+    "true",
+    # let's add some common JSON string values that should not be translated
```
Opoku-Agyemang, line 20: add empty string here

src/_main_/utils/translation/__init__.py
```diff
@@ -45,11 +71,15 @@ def _should_exclude(self, key, _value):
...
             return True
-        
+
         # Check if value is not a string.  For eg. we want to exclude types like bool, int etc
-        if not isinstance(_value, str) :
+        if not isinstance(_value, str):
+            return True
+
+        # Check if value is in JSON_EXCLUDE_VALUES
+        if _value in JSON_EXCLUDE_VALUES or any(re.fullmatch(pattern, _value) for pattern in EXCLUDED_JSON_VALUE_PATTERNS):
```
Opoku-Agyemang, line 80: i'm ok with this for now but i am curious about the impact on the runtime.  for large text blocks does the speed decrease because of the regex comparison?

src/_main_/utils/translation/__init__.py
```diff
@@ -106,46 +136,67 @@ def get_flattened_dict(self):
...
-        # Flattened dictionary keys and values
-        keys = []
-        untranslated_text_entries = []
-        for key,value in self._flattened.items():
+
+        keys = [] # Flattened dictionary keys and values
+        untranslated_text_entries = [] # texts to be translated
+        text_hashes = [] # text hashes to be translated
+
+        for key, value in self._flattened.items():
```
Opoku-Agyemang, line 153: add a TODO: here to parallelize this.  maybe split the list into maybe 10 and have the job of the translation done in several threads and the results combined.

src/_main_/utils/translation/__init__.py
```diff
@@ -106,46 +136,67 @@ def get_flattened_dict(self):
...
+            # Check if the value should be excluded from translation
+            text_hash_exists, err = self.textHasService.text_hash_exists(value)
+            if err:
+                log.error("Error checking if text hash exists for", value, ":", err)
+                continue
+
+            if text_hash_exists:
+                continue
+
+            text_hash, err = self.textHasService.create_text_hash(args = { "text": value })
```
Opoku-Agyemang, line 166: don't we rather want to finish translating then we create the entries for the hash and the translations?  In fact.  can we fan out another thread to do this and not have it dependent on this main thread?  

Actually maybe a better idea is to add it to a queue again to do the processing.  so we don't have to wait for a response here.  the worst case is that we don't store the hash and so we have to translate it again.  so we should not block on the result from this.

Opoku-Agyemang (changes requested):
Great additions!  Left a few nit comments.
I am also not seeing any tests for this.

Opoku-Agyemang (changes requested):

src/_main_/utils/translation/__init__.py
archx3: I will do some reading and run some experiments with fairly large bodies of text to see how it fare and revert

archx3: From my readings on similar issues and a Stackoverflow, [here](https://stackoverflow.com/questions/42742810/speed-up-millions-of-regex-replacements-in-python-3), pre-compiling the regex patterns dramatically improves the performance and good for our use-case.

Opoku-Agyemang (approved):

PR #1122, opened 2024-07-24
Add caching and hash services for translations

archx3:
####  Summary / Highlights
Introduce `TranslationsCacheStore`, `TextHashStore`, and corresponding services to manage caching of translations and text hashing in the database. Also included utility functions for language code handling and text hash creation.

Addresses #1121 

* [ ] I've read and understood the [Contributing Guidelines](/blob/main/.github/contributing.md) and have done my best effort to follow them.


src/api/services/text_hash.py
```diff
@@ -0,0 +1,62 @@
...
+        if not text and not isinstance(text, str):
+            return None, CustomMassenergizeError("Please provide a valid text")
+
+        hash = TextHashService.make_hash(text)
+        args[ 'hash' ] = hash
+
+        text_hash, err = self.store.create_text_hash(args)
+
+        if err:
+            print("CTxH ERR", err)
```
Opoku-Agyemang, line 27: please log instead of print

src/api/services/text_hash.py
```diff
@@ -0,0 +1,62 @@
...
+
+    @staticmethod
+    def make_hash (text: str):
+        return hashlib.sha256(text.encode()).hexdigest()
+
+    def create_text_hash (self, args):
+        text = args.get('text', None)
+
+        if not text and not isinstance(text, str):
+            return None, CustomMassenergizeError("Please provide a valid text")
```
Opoku-Agyemang, line 19: centralize your error messages

src/api/services/translations_cache.py
```diff
@@ -0,0 +1,128 @@
...
+            if not translation_meta:
+                continue
+
+            translatable_fields = getattr(translation_meta, "fields_to_translate", None)
+            if len(translatable_fields) > 0:
+                all_records.extend([ model_to_dict(r, fields = translatable_fields) for r in model.objects.all() ])
+
+        return all_records
+
+    def translate_all_models (self, language_code: str) -> Tuple[ dict or None, None ]:
```
Opoku-Agyemang, line 101:
```suggestion
    def translate_all_db_table_contents (self, language_code: str) -> Tuple[ dict or None, None ]:
```

src/api/services/translations_cache.py
```diff
@@ -0,0 +1,128 @@
...
+                translation, err = self.create_translation( {
+                    "hash": hashes[i],
+                    "source_language": DEFAULT_SOURCE_LANGUAGE_CODE,
+                    "target_language": language_code,
+                    "translated_text": translations[i]
+                })
+
+                if err:
+                    log.error(f"Error caching translation for {hashes[i]}: {err}")
+                    return None, err
```
Opoku-Agyemang, line 122: i like the idea of this.  What do you think about putting a TODO comment here to eventually make this process async.  For instance: we can put all the things to translate in an SQS queue.  then we have something that polls the queue and processes anything that is in that queue by performing this action you have here.  If the creation of the translation fails, we will add it to a dead letter queue so we can track all the insertions that didnt work and can alert on it.  And runs some things manually if need be.  For now this is ok but let's leave a comment to do that in the near future.  I am also open to ideas here.

That said if an error occurred in calling ` self.create_translation` lets not return an error.  Just because one failed does not mean we stop everything.  Maybe the next one will succeed.  So at the very least we should rather be collecting the errors and maybe log some message that X errors occurred.  This is where having that queue would be helpful in the future because if at first it does not succeed, the system will auto retry at some other point.

src/api/store/text_hash.py
```diff
@@ -0,0 +1,51 @@
...
+
+class TextHashStore:
+    def __init__(self):
+        self.name = "Text Hash Store/DB"
+
+    def get_text_hash_info (self, args) -> Tuple[ dict, any ]:
+        try:
+            hash = args.get('hash', None)
+            if not hash:
+                return None, CustomMassenergizeError("Invalid hash provided")
```
Opoku-Agyemang, line 14: centralize all your errors

Opoku-Agyemang (changes requested):
Nice work!  Left some comments

src/api/services/translations_cache.py
archx3: I've rewritten this function to be significantly more robust than the previous one, added retry logic as well as changed the processing by using a deque queue instead of the list.
A pseudo dead letter queue has been added currently for error reporting and also to make it easy to leverage SQS when we're ready.

archx3: SIDE NOTE @Opoku-Agyemang :
I rebased main into this to get the models in. Now the files have increased in number but focus on:
`src/api/services/translations_cache.py`

I made the changes requested in `src/api/services/text_hash.py` and `src/api/store/text_hash.py` but also marked them as deprecated with respect to our last discussion

Opoku-Agyemang (approved):

PR #1132, opened 2024-08-03
Hotfix: Enhance is_value function conditional checks

abdullai-t:
####  Summary / Highlights
This pull request fixes a bug in the `is_value` function. The issue was caused by `actions.update` passing `calculator_action` as an argument, which should be an `int`, but was being passed as `null`. Since `is_value` wasn't checking for `null`, the field validator threw an error.

To resolve this, two new value checks `(null and '')` have been added to the `is_value` function. Additionally, the code has been refactored for improved readability.

* [ ] I linked this PR to the relevant issue.

### DIFF
--- a/src/_main_/utils/common.py
+++ b/src/_main_/utils/common.py
@@ -261,10 +261,15 @@ def extract_location(args):
     return location
 
 
+# def is_value(b):
+#     if b and b != "undefined" and b != "NONE":
+#         return True
+#     if b == "":  # an empty string is a string value
+#         return True
+#     return False
+
 def is_value(b):
-    if b and b != "undefined" and b != "NONE":
-        return True
-    if b == "":  # an empty string is a string value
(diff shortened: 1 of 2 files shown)


Opoku-Agyemang (approved):

src/_main_/utils/common.py
```diff
@@ -261,10 +261,15 @@ def extract_location(args):
...
+#         return True
+#     if b == "":  # an empty string is a string value
+#         return True
+#     return False
+
 def is_value(b):
-    if b and b != "undefined" and b != "NONE":
-        return True
-    if b == "":  # an empty string is a string value
+    if b and b not in ["undefined", "null", "None", ""]:
```
BradHN1, line 272: @abdullai-t - this is not correct.  See the comment above, where "" should return True.  Also note that the empty string tests as b=false, so maybe your change doesn't break anything but it doesn't do as you intend.

PR #1179, opened 2024-10-15
✨ Fix Testimonial Auto-Share Sequence During Creation

abdullai-t:
####  Summary / Highlights
This PR addresses an issue where testimonials were being shared with communities prematurely during their creation process. The function used to fetch communities for auto-sharing was called before assigning the community to the testimonial.

### DIFF
--- a/src/api/store/testimonial.py
+++ b/src/api/store/testimonial.py
@@ -197,11 +197,6 @@ class TestimonialStore:
         user = UserProfile.objects.filter(email=user_email).first()
         if user:
           new_testimonial.user = user
-          
-      if is_published:
-          new_testimonial.published_at = parse_datetime_to_aware()
-          add_auto_shared_communities_to_testimonial(new_testimonial)
-
 
       if action:
         testimonial_action = Action.objects.get(id=action)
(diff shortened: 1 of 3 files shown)


src/api/store/testimonial.py
```diff
@@ -242,6 +237,11 @@ def create_testimonial(self, context: Context, args) -> Tuple[dict, MassEnergize
 
       new_testimonial.save()
       
+      if is_published:
+          new_testimonial.published_at = parse_datetime_to_aware()
+          add_auto_shared_communities_to_testimonial(new_testimonial)
+      new_testimonial.save()
```
HeyMrQuaidoo, line 243: In the code below, we update the `new_testimonial` variable and save it. But after setting the audience, shouldn’t we save it again to make sure the changes to `new_testimonial` are kept?. Thoughts cc @abdullai-t 

```
      new_testimonial.save()

      if audience:
        new_testimonial.audience.set(audience)
```

HeyMrQuaidoo (changes requested):
@abdullai-t Could you elaborate why we don't need to save the new_testimonial variable after setting the audience?

src/api/store/testimonial.py
abdullai-t: The `audience` field in the `Testimonial` model is a ManyToMany relationship. Django automatically creates a bridge table to store data for such fields. Using `.set()` or `.add()` on a ManyToMany field, like in `new_testimonial.audience.set(audience)`, saves the related instances without requiring an explicit `.save()` call on the `Testimonial` instance.

HeyMrQuaidoo (approved):
LGTM

